PR #2, opened 2020-06-01
(dustsettle) corrected dust fraction if a size distribution of grains

jameswurster:


danieljprice:
there is something wrong with my [bots] GitHub action, but aside from that all looks good. Eventually I would like to run the full nightly test suite on pull requests and only merge if things pass.

PR #7, opened 2020-06-03
(divB cleaning) dtclean only divided by 2 if cleaning is required

jameswurster:


danieljprice:
mpi is failing...

jameswurster:
I'll look into this later today... I have a couple of pressing tasks that I need to work on.
Off hand, I'm not sure how these changes would affect MPI since it all changes should be local to a single node...

danieljprice:
I'll try to just run the checks again, I recall the MPI code failing very occasionally with this error in the nightly build, but seemed sporadic

jameswurster:
With how the bots work with pull requests, does this mean that it's definitely an issue with this pull?   I see nothing in this commit that should cause a bug in linear momentum....

jameswurster:
After my new commits (that were not attempting to bug fix the linear momentum issue), the MPI test now passes on Sedov's linear momentum but fails on dE/dt in the Radiation module.

This pull request was open, and when I committed new changes to my fork, they also automatically got added to this pull request.  Is that how this is supposed to work, or did I make an error in my commits?

danieljprice:
yep, I think the pull request updates automatically. But we're still failing MPI here.

jameswurster:
I've been looking at the code, and I don't see what could be causing the mpi fail.  My thought at the moment is to reset the dtclean to be 1/2 as small and see what that does.  However, I found a few things when looking through and testing the code:
-on one mpi test, it failed in the previous radiation test saying dE/dt /= 0 and passed the linear momentum test of sedov
-when running locally without mpi, the Sedov test crashes when performed alone due to dtrad going too small, but passes when run in the full testsuite
-likely a trivial formatting issue, but why does energies use reduce_fn rather than reduceall_mpi?  I see that the two are equivalent, so it is just a naming issue.

I'll keep looking for the bug as I get time.
Cheers
James

________________________________

PR #14, opened 2020-06-21
Add companion gravity

themikelau:


danieljprice:
@themikelau your routine for "get_softened_force" should not hard-wire the softening kernel, you should instead use the appropriate function from kernels.f90

PR #23, opened 2020-07-10
Updates to asteroidwind routines

becnealon:
Changes to setup_asteroidwind and inject_asteroidwind:
- added iexternalforce as option in setup
- fixed constant mass injection rate so that the injection rate requested is correctly recovered
- added two variable mass injection rates varying with
i) cos(t) with most particles injected at pericentre and
ii) r^(-2) with most particles injected at pericentre
- easy to put in other functions for mass inject rate
- added a check for silly variables in setup

### DIFF
--- a/.github/workflows/bots.yml
+++ b/.github/workflows/bots.yml
@@ -24,9 +24,17 @@ jobs:
 
     - name: Run the bots
       run: |
-        git config --global user.email "daniel.price@monash.edu"
-        git config --global user.name "Daniel Price"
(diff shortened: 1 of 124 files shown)


danieljprice:
thanks Bec. Only comment is that it would be nice to shift the helper routines into a utilities module, so they can be reused elsewhere. For example, feel free to change the get_E routine in set_binary if you were having problems with it. Similarly with your integration routines.

PR #35, opened 2020-07-23
setup_dustsettle now reads/writes setup files; added more prompts

jameswurster:


danieljprice:
please note we still have a test failure, but I think this should be safe to merge

PR #36, opened 2020-07-24
formatting changes

jameswurster:


src/setup/setup_dustsettle.f90
```diff
@@ -266,11 +267,22 @@ subroutine setpart(id,npart,npartoftype,xyzh,massoftype,vxyzu,polyk,gamma,hfact,
...
     print*,' particle mass         = ',massoftype(itype)*umass,'g'
     print*,' total mass            = ',npart*massoftype(itype)*umass,'g'
     print*,' mid-plane gas density = ',rhozero*umass/udist**3,'g/cm^3'
+    print*,' Rdisc                 = ',Rdisc*udist/au,'au'
+    print*,' H0                    = ',H0*udist/au,'au'
+    print*,' H/R                   = ',HonR
+    print*,' cs                    = ',cs*udist/utime,'cm/s'
+    print*,' T                     = ',gmw*mass_proton_cgs*(cs*udist/utime)**2/(gamma*kboltz),'K'
+ endif
+ if (gmw*mass_proton_cgs*(cs*udist/utime)**2/(gamma*kboltz) > 20.) then
```
danieljprice, line 281: Couldn't this just be a check on the HonR value? Converting manually to temperature here seems rather inelegant

src/main/readwrite_dumps.F90
```diff
@@ -1077,7 +1077,7 @@ subroutine read_smalldump(dumpfile,tfile,hfactfile,idisk1,iprint,id,nprocs,ierr,
 #ifdef INJECT_PARTICLES
  call allocate_memory(maxp_hard)
 #else
- call allocate_memory(int(nparttot / nprocs) + 1)
+ call allocate_memory(int(1.25*nparttot / nprocs) + 1)
```
danieljprice, line 1080: my feeling is that this is still dangerous, one could easily obtain an imbalance that is more than 25% in the tree build. I would suggest at least a factor of 2 here if nprocs is > 1

jameswurster: The factor of 1.25 was somewhat randomly chosen.  I have increased this to 2 for the moment.  Please see the MPI slack channel for further mpi comments

src/setup/setup_dustsettle.f90
jameswurster: The temperature was initially added since it was a more familiar variable to me.  The warning was then subsequently added.  I have removed both in my subsequent commit.

PR #41, opened 2020-07-29
Splitting/merging

becnealon:
### Splitting/merging
- reformatted routines in splitpart.f90 so they could be more easily called from routines outside of moddump
- fixed small bug in sample_kernel
- added shifting routine for particles (may be important when doing splitting in live calculation)
- added merging routine, formatted the same way as the splitpart utility

### Asteroid wind/injection
- particles are only injected on the side of the asteroid that faces the central potential

### Still to be done
- both merging and splitting routines in splitpart.f90 can definitely be sped up, @jameswurster has already agreed to help with this
- while moddump_mergepart.f90 does work, it currently assigns children by just picking the next closest particle that's not already been picked - there must be a better way to assign children to groups that don't result in the last few children potentially being at large distances?
- neither split or mergeparts deals with energies, don't know if this is important?


src/utils/moddump_splitpart.f90
```diff
@@ -17,20 +17,72 @@ module moddump
...
+ use splitpart,    only:split_particles
+ use io,           only:fatal,error
+ use injectutils,  only:get_parts_per_sphere
+ use part,         only:igas,set_particle_type
  integer, intent(inout) :: npart
  integer, intent(inout) :: npartoftype(:)
  real,    intent(inout) :: massoftype(:)
  real,    intent(inout) :: xyzh(:,:),vxyzu(:,:)
- integer :: ierr
+ real, allocatable, dimension(:,:) :: xyzh_child,vxyzu_child
```
danieljprice, line 35: these are small enough to not need to be allocatable, they are only 4*nchild

src/utils/moddump_mergepart.f90
```diff
@@ -0,0 +1,153 @@
...
+ real,    intent(inout) :: xyzh(:,:),vxyzu(:,:)
+ integer :: ierr,nactive,nparent,remainder, i,j,k
+ integer :: on_list(npart), children_list(nchild)
+ integer :: ichild,iparent,child_found
+ logical :: iactive(npart)
+ real, allocatable, dimension(:,:) :: xyzh_parent,vxyzu_parent,xyzh_in,vxyzu_in
+ real    :: rik(3),rik2,rikmax
+
+ierr = 0
+
```
danieljprice, line 46: general comment here is that there is a lot of functionality dumped into modify_dump, which is not very helpful if you wanted to do this as part of a live calculation, would be better to put all of this into a utility routine

src/utils/moddump_mergepart.f90
```diff
@@ -0,0 +1,153 @@
...
+ use part,      only:igas,kill_particle,delete_dead_or_accreted_particles
+ use io,        only:fatal,error
+ integer, intent(inout) :: npart
+ integer, intent(inout) :: npartoftype(:)
+ real,    intent(inout) :: massoftype(:)
+ real,    intent(inout) :: xyzh(:,:),vxyzu(:,:)
+ integer :: ierr,nactive,nparent,remainder, i,j,k
+ integer :: on_list(npart), children_list(nchild)
+ integer :: ichild,iparent,child_found
+ logical :: iactive(npart)
```
danieljprice, line 41: this is terrible, you cannot allocate a large temporary array of length npart here

src/utils/moddump_mergepart.f90
```diff
@@ -0,0 +1,153 @@
...
+ !-- how many active particles (this routine currently ignores inactive particles)
+ nactive = 0
+ iactive = .false.
+ do i=1,npart
+   if (xyzh(4,i) > 0.) then
+     nactive = nactive + 1
+     iactive(i) = .true.
+   endif
+ enddo
+ if (nactive < npart) print*,'Discarding inactive particles'
```
danieljprice, line 56: suggestion here is just to pull nactive from global memory in the utils_indtimesteps module, the code above is unsafe. You should not rely on xyzh(4,i) > 0 as flag for active/inactive, plus temporarily allocating a large array (iactive) is not nice

src/utils/splitpart.f90
```diff
@@ -26,70 +25,44 @@ module splitpart
 !  split particles into nchild particle
 !+
 !--------------------------------------------------------------------------
-subroutine split_particles(nchild,npart,npartoftype,xyzh,massoftype,ierr)
- use injectutils, only:get_parts_per_sphere
+subroutine split_particles(nchild,iparent,xyzh_parent,vxyzu_parent, &
+           lattice_type,ires,xyzh_child,vxyzu_child)
  use icosahedron, only:pixel2vector,compute_corners,compute_matrices
- use part,        only:copy_particle
```
danieljprice, line 32: the bit I'm uncomfortable with here is that you have deleted use of the copy_particle routine. So if a particle has any properties beyond velocity they will not be inherited by the children, which is extremely dangerous.

src/utils/splitpart.f90
```diff
@@ -99,12 +72,106 @@ subroutine sample_kernel(iseed,dx)
...
- do while (r > 2.)
+ do while (abs(r) > 2.)
     r = gauss_random(iseed)
  enddo
  dx = get_random_pos_on_sphere(iseed)
  dx = r*dx
 
 end subroutine sample_kernel
 
+subroutine shift_particles(npart,xyzh,vxyzu,deltat,beta,shifts)
```
danieljprice, line 83: can you give a comment stating the purpose of this routine?
hint: if you give the comment just after the subroutine statement it will appear in the auto-documentation

src/utils/splitpart.f90
```diff
@@ -99,12 +72,106 @@ subroutine sample_kernel(iseed,dx)
...
+ xyzh_parent = 0.
+ vxyzu_parent = 0.
+ do i=1,nchild
+   ichild = children_list(i)
+   xyzh_parent(1:3) = xyzh_parent(1:3) + xyzh(1:3,ichild)
+   vxyzu_parent(:)  = vxyzu_parent(:) + vxyzu(1:3,ichild)
+ enddo
+ xyzh_parent = xyzh_parent/nchild
+ vxyzu_parent = vxyzu_parent/nchild
+
```
danieljprice, line 152: needs a copy particle here for other particle properties

src/utils/moddump_mergepart.f90
```diff
@@ -0,0 +1,153 @@
...
+  elseif (remainder > 0) then
+    ! ignore a couple
+    do i = 1,remainder
+      if (xyzh(4,npart-i) > 0.) iactive(npart-i) = .false.
+    enddo
+    print*,'Ignoring',remainder,'particles to get an even split'
+ endif
+
+ !-- arrays for new parent particles
+ allocate(xyzh_parent(4,nparent),vxyzu_parent(3,nparent))
```
danieljprice, line 75: this will not work if particles have any other properties beyond velocity and u. The best way to store temporary particles would be to temporarily store them at the end of the real particle list, using copy_particle to copy the properties. Just copying position and velocity is dangerous.

src/utils/splitpart.f90
```diff
@@ -26,70 +25,44 @@ module splitpart
...
+     call pixel2vector(j-2,ires,geodesic_R,geodesic_v,dx)
+   else
+     call sample_kernel(iseed,dx)
+   endif
+      xyzh_child(1:3,j) = xyzh_parent(1:3) + sep*dx(:)
+      vxyzu_child(1:3,j) = vxyzu_parent(1:3)
  enddo
 
- !--find positions of the new particles
- inew = npart ! put new particles after original parts
```
danieljprice, line 74: the new particles being placed at the end of the list was important, because it allows the use of the copy_particle routine to copy the properties

danieljprice:
Thanks for the pull request - looks like excellent progress here. I dropped a "review" above on some of the things that are unsafe in the code, but in general it's very nice to have the ability to merge particles.

src/utils/moddump_splitpart.f90
becnealon: nchild can change depending on whether you're using a grid or the random arrangement for the children particles

src/utils/moddump_mergepart.f90
becnealon: Sure - I tried to arrange them so that the moddumps contained everything related to the fact that it's over all particles, rather than just a subset that would be used in a live calculation. Can rearrange it back?

src/utils/moddump_mergepart.f90
becnealon: 😆  this is because I couldn't work out how to keep track of which particles had been paired and which hadn't. I think James said he'd help me with this next week

src/utils/moddump_mergepart.f90
danieljprice: I would suggest just moving the stuff in moddump to a new routine, then in principle one could call either the "split one particle" routine and the "split all particles" routine

src/utils/moddump_splitpart.f90
danieljprice: see comment below, if you make the stuff in moddump into a subroutine, you could pass nchild as as argument, then you can simply declare real, xyzh_child(4,nchild), vxyzu_child(4,nchild). I think this is small enough not to cause stack problems.

src/utils/moddump_mergepart.f90
danieljprice: nothing wrong with the overall structure, just would be nice to have basically nothing in the moddump itself

danieljprice:
Just an additional comment here. One other (possibly terrible) way to merge particles would be to just roll the dice for the probability of converting a low-mass particle back to a high mass particle. If the dice-roll fails, the low-mass particle just gets deleted... The advantage of this is that it would be super-simple, disadvantage is that it is obviously noisy. But avoids the problem of "what if I don't have 13 particles to merge?"

src/utils/moddump_mergepart.f90
becnealon: nactive doesn't appear to have been set when running this as a moddump. I tried calling set_active_particles to set it but then I had to allocate iphase and ibin and I ran into errors with ibini and nbinmax, defeating the purpose. Changed it so that at the start it uses isdead_or_accreted for each particle, deletes all the inactive particles and then carries on only using active particles

src/utils/splitpart.f90
becnealon: forgot, fixed

becnealon:
How do you evaluate the probability of whether a particle should merge or not? I don't think I understand this point fully, but don't like the idea of ignoring more than a handful of particles and it sounds like you might be discarding quite a few ...

becnealon:
I think most of your above comments are fixed by putting back in copy_particle and copying them in the same way for both merging and splitting (after npart). Reformatted so it can be called on either a subset or all of the particles more easily, with the intention that only the utils will be needed in live calculations.

src/utils/moddump_mergepart.f90
becnealon: Will still work to fix this, I don't think this is the best way to arrange the children

danieljprice:
can this be merged?

becnealon:
not yet please

becnealon:
Done and ready to merge, thanks

danieljprice:
this looks great now, thanks @becnealon and @jameswurster

PR #49, opened 2020-08-11
bug fixes to radiation hydro, also implemented radiation pressure

emborchert:


danieljprice:
this looks great so far. The important changes are:
- pressure and sound speed are now stored for each particle
- the equation of state is called ~100 times less often, this should result in a speedup for complicated equations of state (e.g. MESA, Helmholtz)
- a bunch of ugly ifdefs have been removed
- the structure of the non-relativistic code matches the relativistic code, in particular having a single place where the "conservative to primitive" transformation takes place (for non-rel code this is just B/rho->rho, sqrt(rho*eps)->eps)

Remaining issues before it can be merged are:
- we broke something in the dust evolution terms, failing "make test", likely from shifting dustevol->dustfrac transformation inside the cons2prim routine
- sedov test is running with very short timestep with radiation switched on, this may just be a units issue, but may also indicate a bug with the radiation pressure/pdV work term in the radiation energy

danieljprice:
- dust evolution bug is fixed, the problem with the sedov test is the main one left

Issue #51, opened 2020-08-13
Interactivity in phantomsetup


jameswurster:
In many of the setup routines, phantomsetup exits after asking the interactive questions.   This is to ensure that the setups are written in such a way that the .in and .setup files can be passed to other researchers and the will be able to reproduce the results.  However, there are some setups (e.g. setup_shock or setup_sphereinbox, and possibly others) where the interactive part of the code either asks about values to be written to the .in file, or calls part of the code that defines values for the .in file.    This results in many of the values in .in being the code default values rather than values appropriate for the given setup.  

Therefore, an elegant solution must be determined, and then several of these setup files will need to be modified; I am only aware of setup_shock and setup_sphereinbox, but please add to this list as appropriate.

danieljprice:
for the setup_disc routines, the current solution is that a parameter determining tmax is read from the .setup file (i.e. end time in number of orbits), and written to the .in file as tmax.

danieljprice:
I just merged this, adding tmax and dtmax to the .setup file is fine, but I don't understand why we have things like alpha, alphau, nfulldump added to the .setup files -- why are these necessary?

PR #54, opened 2020-08-14
added values to .setup as per issue #51

jameswurster:
As discussed, I have added the values relevant to the .in file and dumpfile into the .setup files for sphereinbox and shock.
In the testsuite, I have also reset iverbose between every test since that value is changed within a few specific tests.


danieljprice:
this all looks good, but just a heads up I am going to delay merging for 1-2 days as I would like to finally get master passing the nightly build (this has been broken since July) before merging anything else on top of it.

PR #56, opened 2020-08-18
(setup_shock) removed some parameters out of the .setup file

jameswurster:
In my previous commit (pull request #54), I added all .in file parameters that were in the choose_shock subroutine to the .setup file.  Here, I removed the parameters that are constant for all shocks, and these parameters are now defined in the main setup subroutine.  Although alphau differs between shock 10 and the rest, this value is initialised differently in options.f90 for gr=.true. which is required for shock 10.  alpha remains in the .setup since it is either 0 or 1, depending on the shock choice.  The remaining .in parameters in the .setup file vary depending on the shock.

### DIFF
--- a/src/setup/setup_shock.F90
+++ b/src/setup/setup_shock.F90
@@ -28,13 +28,12 @@ module setup
 !   set_dust, setshock, setup_params, timestep, unifdis, units
 !
  use dim,       only:maxvxyzu,use_dust,do_radiation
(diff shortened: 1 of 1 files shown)


danieljprice (commented):
thanks. We can remove the explicit setting of alphau for the relativistic shock, this should be done via the .in file, there is no need to enforce it in the .setup file

jameswurster:
alphau is not set in setup_shock; it remains as a comment to show that it is different for shock 10 compared to the other shocks.

PR #71, opened 2020-09-11
ptmass: sink output files

jameswurster:
For enhanced performance, Phantom will now write the sink properties (e.g. mass, position, velocity, etc) to a single ascii file rather than nptmass separate files.  This should enhance performance for large nptmass since we only need to access and flush one file instead of nptmass.  This will also make the output folder cleaner by not generating so many files.

I have added a program that will post-process this single sink file into nptmass files. This has the added benefit over the original version of having one file per sink rather than one file per sink per restart.  compile and execute using
make sinks; ./phantomsinks


src/utils/phantomextractsinks.f90
```diff
@@ -0,0 +1,104 @@
...
+    stop
+ endif
+ !
+ !--Determine the first time in each file
+ !
+ tstart = huge(tstart)
+ do i = 1,nfiles
+    write(filename,'(a,I2.2,a)') trim(prefix),i,'.sink'
+    open(unit=20,file=trim(filename))
+    read(20,'(a)') cdummy
```
danieljprice, line 67: if the intention here is just to skip a line, you can just use 
```
read(20,*)
```
i.e. there is no need to read a long string

src/setup/set_sphere.f90
```diff
@@ -235,10 +237,12 @@ subroutine set_unifdis_sphereN(lattice,id,master,xmin,xmax,ymin,ymax,zmin,zmax,p
  real,             intent(out)   :: psep,xyzh(:,:)
  integer(kind=8),  intent(inout) :: npart_total
  integer,          intent(out)   :: ierr
+ integer,          parameter     :: itermax = 100
  procedure(mask_prototype)       :: mask
  integer(kind=8)                 :: npart_local
- integer                         :: nps_lo,nps_hi,npr_lo,npr_hi,test_region,iter,nold(4)
+ integer                         :: i,iter,test_region,nps_lo,nps_hi,npr_lo,npr_hi
  integer                         :: npin,npmax,npart0,nx,np,dn
+ integer                         :: ninout(2,itermax),nold(4)
```
danieljprice, line 245: can the same be achieved here without storing a large array? These kind of large temporary arrays cause seg faults.

src/setup/set_sphere.f90
```diff
@@ -336,15 +342,23 @@ subroutine set_unifdis_sphereN(lattice,id,master,xmin,xmax,ymin,ymax,zmin,zmax,p
        return
     endif
     ! always use more particles than requested
-    npart = nps_hi
+    npart_total = npart_local
+    npart = npin
+    do i = 1,iter
+       if (ninout(2,i)==nps_hi) nx = ninout(1,i)
```
danieljprice, line 348: could this be computed on the fly without having to store the result for every iteration?

danieljprice (commented):
Hi @jameswurster. Just curious - was this from an actual or perceived performance issue? Because the number of files is in the end the same (actually worse, because after running phantomsinks we have two copies of all the sink information).

I wonder if this might be better as an option rather than the default, since it adds an extra hoop to jump through for the user. For example, maybe should do this if the number of sink particles > 10 (or if icreate_sinks is on?), but could stick with the current behaviour for simulations with < 10 sink particles?

I am thinking for things like circumbinary disc simulations, it is very helpful to have the file separately for each sink

src/utils/phantomextractsinks.f90
jameswurster: Done

src/setup/set_sphere.f90
jameswurster: Done.
The array was only 2x100, so not memory intensive at all.

src/utils/phantomextractsinks.f90
jameswurster: The sink file change was for a few reasons, both practical and perceived.  
I'm running converging flow simulations where several dozen sinks can form.  This naturally leads to many small files, especially if restarts are required; this makes the folder unbearably messy, especially if several simulations are run in the same folder.
It is computationally more efficient to write & flush one file than a few dozen files.
I'm applying for time on Dirac, and they explicitly want to minimise the number of small files that are generated and accessed.  

It was for all the above reasons that I made this change.  Although my change requires one extra step, the post-processor has the added benefit of generating files that do not duplicate data if a restart does not occur at the exact simulation time that the code ended.

I understand the simplicity of the former method, especially if there is only a few sinks. If you would like, I can re-add the previous method and let the user chose between outputs with (e.g.) a flag in the .in file (leaving the default at the original).   I think in the long run, the new method may be more efficient at runtime, especially if we switch to writing the sink files as unformatted files.

danieljprice: I don't want to put any more options in the .in file, but I think we should apply a threshold here. A simple one would be just to use the "single file" method if create_sinks = .true., but print one file per sink otherwise.

jameswurster: Sounds good.  I'll make this change and then re-commit.

jameswurster: I have readded the old method, to be used if icreate_sinks == 0; the new method with be used if icreate_sinks > 0.

PR #74, opened 2020-09-22
Implemented GR BH-NS ejecta setup and r-process heating function

sdarbha:
In this branch, I implemented (1) the setup for a BH-NS tidal ejecta in GR and (2) an r-process heating function. The aim of this pull request is to compare the code, not to actually merge the branch.


build/Makefile
```diff
@@ -272,6 +272,23 @@ ifeq ($(SETUP), grtde)
     ANALYSIS=analysis_tde.f90
 endif
```
danieljprice, line 274: looks good to me, feel free to delete the commented-out lines

build/Makefile
```diff
@@ -1223,6 +1241,33 @@ ifeq ($(SYSTEM), g2)
...
 
+# I (SD) added this system to be able to compile Phantom on NERSC Cori
+# On Cori, you load programming environments, which use wrappers for the compilers, you do not use the vendor specific compiler names directly
+# I (SD) just copied the parameters from Makefile_defaults_ifort and ozstar below and made several replacements
+# Most importantly, I (SD) made the replacements ifort -> ftn and icc -> cc, and added CRAYPE_LINK_TYPE to tell Cori how to link
+ifeq ($(SYSTEM), cori)
+#Cori (NERSC machine)
+    FC= ftn
+#     FFLAGS= -O3 -inline-factor=500 -mcmodel=medium -shared-intel -warn uninitialized -warn unused -warn truncated_source
+    FFLAGS= -O3 -inline-factor=500 -shared-intel -warn uninitialized -warn unused -warn truncated_source
```
danieljprice, line 1252: here just use "include Makefile_defaults_ifort" and then include only any lines which differ from the default settings from the ifort compiler (e.g. FC). Please delete the lines which don't change (e.g. DBLFLAG, DEBUGFLAG, ENDIANFLAG etc)

src/main/cooling.F90
```diff
@@ -396,12 +410,15 @@ end subroutine implicit_cooling
...
 !-----------------------------------------------------------------------
-subroutine energ_cooling (xi, yi, zi, u, dudt, rho, dt, Trad, mu_in, K2, kappa)
- real, intent(in) :: xi, yi, zi, u, rho, dt !in code unit
+subroutine energ_cooling (xi, yi, zi, u, dudt, rho, dt, Trad, mu_in, K2, kappa, t)
+ real, intent(in) :: xi, yi, zi, u, rho, dt ! in code unit
  real, intent(in), optional :: Trad, mu_in, K2, kappa ! in cgs!!!
+ real, intent(in), optional :: t ! in code units
  real, intent(inout) :: dudt
 
  select case (icooling)
```
danieljprice, line 419: looks good to me

src/main/rprocess_heating.f90
```diff
@@ -0,0 +1,179 @@
+!--------------------------------------------------------------------------!
+! The Phantom Smoothed Particle Hydrodynamics code, by Daniel Price et al. !
```
danieljprice, line 2: this module looks beautiful

src/main/step_leapfrog.F90
```diff
@@ -812,6 +826,23 @@ subroutine step_extern_gr(npart,ntypes,dtsph,dtextforce,xyzh,vxyzu,pxyzu,dens,me
 
           pxyz      = pxyz + hdt*fexti
```
danieljprice, line 828: this also seems ok, by analogy with the non-relativistic version. However, I think the call here could just be general, i.e. if (icooling > 0). There is no reason why the other cooling prescriptions could not also be used in the relativistic code

src/utils/analysis_bhnsejecta.f90
```diff
@@ -0,0 +1,180 @@
...
+ real    :: m, densi
+ real    :: T, s, y_e
+ real    :: q
+ real    :: time_output, time_output_seconds
+ character(len=120) :: particle_file_name, info_file_name
+
+
+
+ particle_file_name = trim(dumpfile)//'.dat'
+
```
danieljprice, line 55: would suggest using open(newunit=iunit) here rather than hardwiring unit 1, just declare iunit as an integer

src/utils/analysis_bhnsejecta.f90
```diff
@@ -0,0 +1,180 @@
...
+ open(1, file=particle_file_name, action='write', status='replace')
+
+ write(1,'(A)',IOSTAT=iostatus) '# Particle data'
```
danieljprice, line 58: house formatting is no SHOUTING, so "(a)" and "iostat" here

danieljprice (commented):
overall the code looks good. Obviously the build failures in the pipeline will block a merge at the moment but the code all looks fine to me. Thanks for submitting the PR, very helpful to check and review the code here.

danieljprice:
build failures can be checked e.g. with "make testkd" on your own machine, or by looking at the logs for the actions above

PR #80, opened 2020-10-05
Fix art con

themikelau:
This fixes an issue experienced by @miguelglezb when he ran with the adiabatic EoS. When compiling using SETUP=star, store_temperature is .true. by default. But this means the force module will try to calculate the artificial conductivity using dU = c_V dT, even though the temperature is not actually stored with the adiabatic EoS, and so gives rubbish. This temporary fix ensures dU = c_V dT is only performed when using a non-ideal EoS.


src/setup/density_profiles.f90
```diff
@@ -475,8 +475,11 @@ subroutine read_mesa(filepath,rho,r,pres,m,ene,temp,Xfrac,Yfrac)
 
  do i = 1, rows
     if (trim(header(i))=='mass_grams') m(1:lines) = dat(1:lines,i)
+    if (trim(header(i))=='mass') m(1:lines) = dat(1:lines,i) * 1.989e33
```
danieljprice, line 478: what is the unit conversion here? If it is an attempt to convert into code units, then umass should be used, rather than hardwiring an arbitrary number which could be wrong

src/main/force.F90
```diff
@@ -1279,7 +1280,7 @@ subroutine compute_forces(i,iamgasi,iamdusti,xpartveci,hi,hi1,hi21,hi41,gradhi,g
 
        if (iamgasj .and. maxvxyzu >= 4) then
           enj   = vxyzu(4,j)
-          if (store_temperature) then
+          if (store_temperature .and. ((ieos == 10) .or. (ieos == 12))) then
```
danieljprice, line 1283: this is dangerous. There are several other non-ideal EOS routines where we would need to modify the conductivity. It would be better to add a logical query function in eos.f90 like eos_is_non_ideal(ieos) which returns .true. if c_v is not constant, i.e.:
```
logical function eos_is_non_ideal(ceos)
 integer, intent(in) :: ieos

 select case(ieos)
 case(10,12,15...)
     eos_is_non_ideal = .true.
 case default
    eos_is_non_ideal = .false.
 end select

end function eos_is_non_ideal
```
the main point here is to put this logic in eos.f90, rather than distributing it throughout the code. Then if you add another EOS, you only need to change routines in eos.f90

danieljprice (changes requested):

src/setup/density_profiles.f90
themikelau: The subroutine "read_mesa" is only used in the softening procedure, which requires this conversion because all of the profile softening modules assume cgs units. This is therefore independent of what the user chooses as their code units.

src/main/force.F90
themikelau: Agreed. I am happy to add this kind of logic and leave this pull request open for now.

src/setup/density_profiles.f90
danieljprice: ok, still better to use solarm from the physcon module rather than hardwiring the number, then the meaning of the number is clearer

danieljprice: also, 1e33 is above the largest real*4, so you need to use 1.989d33 here

themikelau: Agreed—I will make these changes.

danieljprice:
@themikelau I am merging this as currently this bug causes NaNs when running phantom. But please push a fix to the issues above when you can.

PR #84, opened 2020-10-16
Added additional setup variables for the MESA EoS

robertoiaconi:
Now when setting up a star using the MESA EoS the code asks for mean molecular weight, hydrogen mass fraction and metals mass fraction.


danieljprice:
thanks Roberto!

PR #86, opened 2020-11-10
(eos) temperature is now an output from eos

emborchert:
temperature is always an output from eos, as there was a problem with the option store_temperature of temperatures being overwritten.


danieljprice:
great. This now means that the store_temperature option can be removed from the code. Will raise an issue about this.

PR #89, opened 2020-11-12
updated cooling & elliptical initial conditions

jameswurster:
1) Cooling:
Added Koyama & Inutuska (2002); this and Gammie cooling are solved explicitly in force; the remaining cooling options are still implicitly calculated in step.  I am not confident that the implicit cooling is fully and correctly operational.  I have corrected a few bugs and added a few fail-safes.  Concern 1 is that the cooling timestep was never initialised nor calculated in step (the latter has been corrected).  Concern 2 is that the new default cooling requires a pre-processor flag and optional arguments, which is very dangerous; default cooling should not require pre-processor flags or optional arguments (so I suggest returning the default to Gammie).  

2) Initial conditions:
Added option to initialise an ellipse or an elliptical void; at the moment, we cannot initialise an elliptical shell.

3) Miscellaneous:
There are several small changes, including updating the documentation to reflect GitHub and not BitBucket

### DIFF
--- a/docs/fork.rst
+++ b/docs/fork.rst
@@ -1,6 +1,12 @@
 Forking Phantom
 ===============
 
+Warning
+-------
+
+This instructions on this page are out of date.
+
+
 Why fork?
 ---------
 
--- a/src/main/cooling.F90
+++ b/src/main/cooling.F90
@@ -9,6 +9,8 @@ module cooling
 ! Gas cooling
 !
 ! :References:
+!   Koyama & Inutsuka (2002), ApJL 564, 97-100
+!   Vazquez-Semadeni, et.al (2007), ApJ 657, 870-883
 !   Townsend (2009), ApJS 181, 391-397
 !   Gail & Sedlmayr textbook Physics and chemistry of Circumstellar dust shells
 !
@@ -37,29 +39,39 @@ module cooling
  implicit none
  character(len=*), parameter :: label = 'cooling'
 
- public :: init_cooling,calc_cooling_rate,energ_cooling
+ public :: init_cooling,init_cooling_type,calc_cooling_rate,energ_cooling
  public :: write_options_cooling, read_options_cooling
  public :: find_in_table
  logical, public :: calc_Teq
- real, public:: bowen_Cprime = 3.000d-5
+ logical, public :: cooling_implicit
+ logical, public :: cooling_explicit
+ real,    public :: bowen_Cprime = 3.000d-5
 
  private
  integer, parameter :: nTg = 64
  integer, parameter :: maxt = 1000
  real,    parameter :: Tref = 1.d5, T_floor = 10.
  integer :: nt
- real :: temper(maxt),lambda(maxt),slope(maxt),yfunc(maxt)
- real :: beta_cool  = 3.
- real :: habund     = 0.7
- real :: temp_floor = 1.e4
- real :: Tgrid(nTg)
+ real    :: temper(maxt),lambda(maxt),slope(maxt),yfunc(maxt)
+ real    :: beta_cool  = 3.
+ real    :: habund     = 0.7
+ real    :: temp_floor = 1.e4
+ real    :: Tgrid(nTg)
+ real    :: crate_coef
  integer :: icool_radiation_H0 = 0, icool_relax_Bowen = 0, icool_dust_collision = 0, icool_relax_Stefan = 0
  character(len=120) :: cooltable = 'cooltable.dat'
 
 
 contains
 
+!-----------------------------------------------------------------------
+!+
+!  Initialise cooling
+!+
+!-----------------------------------------------------------------------
 subroutine init_cooling(ierr)
+ use units,   only:utime,umass,udist
+ use physcon, only:mass_proton_cgs
  use io,  only:fatal
  integer, intent(out) :: ierr
 
@@ -83,15 +95,32 @@ subroutine init_cooling(ierr)
 #endif
  calc_Teq = (icool_relax_Bowen == 1) .or. (icool_relax_Stefan == 1) .or. (icool_dust_collision == 1)
 
-!initialize grid temperature
+ !--initialise remaining variables
  if (icooling == 2) then
     call init_cooltable(ierr)
+ elseif (icooling == 5) then
+    crate_coef = 2.0d-26*umass*utime**3/(mass_proton_cgs**2 * udist**5)
  elseif (icooling > 0) then
     call set_Tgrid
  endif
 
 end subroutine init_cooling
 
+subroutine init_cooling_type(h2chemistry,ierr)
+ integer, intent(out) :: ierr
+ logical, intent(in)  :: h2chemistry
+
+ if (h2chemistry) then
+    if (icooling > 0) cooling_implicit = .true.    ! cooling is calculated implicitly in step
+ elseif (icooling > 0) then
+    if (icooling == 3 .or. icooling == 5) then
+       cooling_explicit = .true.                   ! cooling is calculated explicitly in force
+    else
+       cooling_implicit = .true.                   ! cooling is calculated implicitly in step
+    endif
+ endif
+
+end subroutine init_cooling_type
 
 !-----------------------------------------------------------------------
 !+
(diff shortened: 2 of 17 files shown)


danieljprice (commented):
feel free to update them...

docs/fork.rst
```diff
@@ -1,6 +1,12 @@
 Forking Phantom
 ===============
 
+Warning
+-------
+
```
danieljprice, line 6: feel free to update them ;)

src/main/cooling.F90
```diff
@@ -396,20 +443,31 @@ end subroutine implicit_cooling
...
+    else
+       call fatal('energ_cooling','Koyama & Inutuska cooling requires gas temperature')
+    endif
  case default
     !call exact_cooling(u, dudt, rho, dt, Trad, mu_in, K2, kappa)
     !call implicit_cooling(u, dudt, rho, dt, Trad, mu_in, K2, kappa)
-    call explicit_cooling(u, dudt, rho, dt, Trad, mu_in, K2, kappa)
+    if (present(Trad) .and. present(mu_in) .and. present(K2) .and. present(kappa)) then
+       call explicit_cooling(ui, dudt, rho, dt, Trad, mu_in, K2, kappa)
+    else
```
danieljprice, line 468: ugh, this is yuck, we need to fix it!

src/main/cooling.F90
```diff
@@ -594,14 +652,14 @@ subroutine write_options_cooling(iunit)
...
  if (h2chemistry) then
     call write_inopt(icooling,'icooling','cooling function (0=off, 1=on)',iunit)
     if (icooling > 0) then
-       call write_inopt(C_cool,'C_cool','factor controlling cooling timestep',iunit)
        call write_options_h2cooling(iunit)
     endif
  else
-    call write_inopt(icooling,'icooling','cooling function (0=off, 1=physics, 2=cooling table, 3=gammie)',iunit)
+    call write_inopt(icooling,'icooling','cooling function (0=off, 1=explicit, 2=Townsend table, 3=Gammie, 5=KI02)',iunit)
     select case(icooling)
```
danieljprice, line 663: this seems to be mixing up the options here. Whether the cooling is performed implicitly or explicitly is a different question to which physical cooling prescription to employ (which is what icooling should denote). Ideally the physical cooling should "just work" without having to specify implicit vs explicit.

danieljprice:
Hi @jameswurster, overall changes looks good, though obviously some things we as a collaboration need to address in the code re: optional arguments to cooling (see comments above). 

I'd also like to understand the reason for wanting to put an option for explicit cooling back into force? Did you see any difference between having cooling done in force and cooling called from the step routine? There should be no difference between the two aside from the possibility of performing substepping in the latter.

If possible would also be good to discuss how we want the "icooling" flag to work. It's a bit of a mash now between the choice of cooling prescription and the choice of algorithm (implicit/explicit), and I feel like we should clearly distinguish the two.

Thanks for the updates to the docs also.

docs/fork.rst
jameswurster: Since forks are mandatory in GitHub, it might be worth just deleting the fork page.

src/main/cooling.F90
jameswurster: The optional arguments were added by you and Seiss.  I just made it safer by adding the if present.  
I agree, though, that this is ugly.  Moreover, this is the default cooling, which is not safe since these optional agruments are only passed in if nucleation = yes.

src/main/cooling.F90
jameswurster: This was a poor renaming on my part.  I dislike how 1 was simply called 'physics', since all options are physically motivated.  I renamed it 'explicit' based upon the subroutine name.  For clarity, this should be something else, but I am unsure what this cooling option actually does or is based upon; since you are more familiar with this modification, it would probably be best for you to rename it.
The decision to implicitly or explicitly calculate cooling is made by Phantom behind the scenes based upon the choice of icooling.

jameswurster:
The optional arguments were initially added by Siess and not me; in a later update, I needed to add the optional argument of Tgas in advance of the removal of the get_temperature function.  As mentioned above, these optional arguments are dangerous given what the default setting of the cooling.  Thus, I agree that something needs to be done.  

The icooling flag works as before; the confusion resulted from a poorly renamed option (based upon a poorly named subroutine from Seiss).  See above comment.

Cooling should *only* be done in force if icooling = 3,5, and should *only* be done in step for the remaining options.

I did notice a difference in my results, which is why I looked into this more and ultimately put Gammie & KI02 heating back.  The discrepancy occurs since heating is calculated in force.  Therefore, the internal energy passed into cooling will be different depending if cooling occurs in force (same energy as used to calculate the heating) or in step (higher energy since the gas has already been heated).  Since cooling is very dependent on temperature, this change in input energy will make a big change in output cooling.  I cannot reconcile to myself why two opposing processes (heating & cooling) should be treated differently with inconsistent input energies.  If you want/need cooling to be done implicitly in step, for consistency, I would argue that heating also needs to be done in step.  
Given the inconsistent treatment of heating/cooling, the bugs that I found when investigating cooling (most of which are fixed, including that there was no cooling timestep in step) and the bugs that remain (e.g. default settings + optional arguments, and that the initial cooling timestep is based upon force & courant and not upon the cooling), I am not fully confident in the current implementation of implicit cooling.  Thus, I have made the above changes and will be be using explicit cooling for my current project.  

I didn't realise the docs were out of date until a student contacted me with issues downloading Phantom...

I would be happy to discuss this in greater detail if you wish.

src/main/cooling.F90
danieljprice: agreed

src/main/cooling.F90
danieljprice: I believe option 1 uses Simon Glover's cooling routines, which has a wide range of physics implemented as described in the phantom paper. This option also does not require any table lookups, so should be safe to use as the default option. It should do something similar to the KI02 cooling when called with similar temperatures/abundance assumptions.

danieljprice:
The main argument for the cooling done in step is that decreasing dtcool below dtcourant does not cost you anything. That is, with cooling in step, if dtcool decreases below dtcourant then the code just automatically takes substeps where only the cooling terms are evaluated on the right hand side (with heating terms fixed). Obviously if this gives problems then it may not be a good idea, but it is a very simple way of accelerating a calculation with fast cooling. Otherwise a small dtcool implies evaluating the full right hand side (density + force + cooling) on the shorter timestep, which is obviously costly.

if possible, please can you post a figure showing how results differ between the two prescriptions? This is mainly for reference in case we revisit the issue.

src/main/cooling.F90
danieljprice: see comment below, I believe the physical cooling routines should not require these optional arguments. I will have a look at this to see what needs to change. I completely agree we need the default cooling option (icooling=1) to "just work" without the code having to be compiled a special way.

jameswurster: For simplicity, I suggest returning the default cooling to Gammie.  It is simple and does not require any optional inputs or additional pre-processor flags.  

I should probably un-commit my change where I added Tgas as an optional input and always recalculate the temperature (this would require keeping the get_temperature function).  This will be necessary if KI02 cooling is ever moved to step since the temperature will be changing on each substep (since vxyzu(4,i) is changing), but the already computed temperature is not (since equationofstate is not being re-called).

jameswurster:
Qualitatively, the results look very similar.  The energies are also indistinguishable unless you zoom in.  
Cooling affects density, as shown in the attached plots (ignore the 'New Phantom' line since it's particles are initialised slightly differently).
oldnew_09rhomax
oldnew_10rhoave

I also found that implicit cooling is slightly slower:
oldnewtime_06t_wallh
In my simulations, there is a reasonable amount of heating, which drives down the timestep.  If cooling is done in force, it counteracts some of the heating so that the timestep is never as small.  For my simulations, cooling never controlled the timestep (but heating did).  

Moreover, shouldn't dtcool always be used when maxvxyzu>=4 (even for implicit cooling) to control the heating timestep?  Unfortunately, my tests of this failed with the Sedov blast wave due to its initially 0 energy for most of the particles.


What changes would you like me to make prior to this commit being merged?
Cheers

Issue #27, opened 2020-07-15
particle ids not tracked in MPI


danieljprice:
_Originally posted to the mailing list, basic issue is that particle IDs are not currently tracked in MPI, this needs implementing:_

I'm trying to plot the trajectory of some gas particles in my phantom simulation. If I understand the code correctly, the indices of the particles serve as the particle IDs, so I could use the index to select a certain particle at different timestep (I think splash does the same way). This works fine if I only use OPENMP parallelization. However, if I use the hybrid MPI+OPENMP parallelization, the trajectory looks quite strange (see the attached plot). I suspect the indices might have been changed while the processors communicate by MPI. I wonder whether this is a bug in Phantom, or it is a feature to accelerate the simulations with MPI enabled. 
If the indices have to be changed with MPI enabled, then I have to switch back to OPENMP, but it is known that OPENMP parallelization is not very scalable in general. For my problem: I am studying the effects of stellar bars/spirals on the galactic gaseous disk, and the bar usually drives a lot of gas into the galactic center, making a high-density ring there which cause the code runs quite slow (takes > 1 month using ~5 million particles with ISOTHERMAL, OPENMP and INDEPENDENT TIMESTEP, CPU: 32 X Intel® Xeon® Gold 6136 processor, Compiler: ifort 18).   I am not sure whether there is a better way to make the code runs faster with or without MPI (e.g. how many threads per physical core is the most optimized configuration in Phantom?).

jameswurster:
With the kill_particle being routinely called (either if r > rkill or to remove dead particles upon restart), particle IDs will now be important for ALL runs.  
While this should be relatively simple to add to Phantom itself, it will take some time, effort and caution before it's successfully added to all the analysis routines and Splash.

jameswurster:
This issue should be solved by commit #96

PR #96, opened 2020-12-04
Tracking particle IDs

jameswurster:
Particle IDs are now tracked.  This is required for MPI and when particles are added/killed between dumps.  In analysis files, to always perform calculations on the same particle, use
do = 1,npart
   j = iorig(i)
  xi = xyzh(1,j)
  ...
enddo

This should solve issue #27 , although some debugging for MPI and hdf5 dumps may be required.

### DIFF
--- a/src/main/part.F90
+++ b/src/main/part.F90
@@ -44,14 +44,15 @@ module part
 !--basic storage needed for read/write of particle data
 !
 
- real, allocatable :: xyzh(:,:)
- real, allocatable :: xyzh_soa(:,:)
- real, allocatable :: vxyzu(:,:)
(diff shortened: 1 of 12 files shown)


danieljprice:
All seems good to me. Does this mean splash automatically rearranges them back into the correct order?

jameswurster:
There currently is no automatic rearranging of particles (for either splash or phantomanalysis).  I'm not sure if this is realistically possible since if particles are deleted, then we would need to create a placeholder for them, which could balloon the array size (especially for splitting).
Since this array is now implemented, we should consider removing accreted particles from the dump files along with the dead particles.
(I just realised there is a bug in my implementation of new IDs, which I will fix shortly)

Issue #91, opened 2020-11-24
.setup file for stars should not ask for mean molecular weight


danieljprice:
The .setup file for SETUP=star currently asks for mean molecular weight as well as composition (X,Z) if the MESA equation of state is employed:

<img width="634" alt="image" src="https://user-images.githubusercontent.com/12252103/100043513-b2b76180-2e61-11eb-8076-5422c4e7181f.png">

There are several issues with this:
- mean molecular weight is an *output* of the MESA equation of state, not an input
- X and Z are over-written anyway in the .in file based on the values at 0.5*Rstar, so they should not need to be asked for in the .setup file

The right procedure should be:
- remove mu, X and Z from the .setup file
- determine mu automatically during the setup and write it to the .in file for adiabatic or ideal+rad equations of state
- if using the MESA EoS, automatically determine the default values of X and Z from the stellar tables (if possible) and write these to the .in file. If X and Z are not found in the stellar tables it should be possible to specify them manually.
- Do NOT write or read mu to the .in file if any non-ideal/tabulated equation of state is employed (because mu is individual to each particle in this case)

danieljprice:
now fixed

PR #98, opened 2020-12-09
Star cleanup

themikelau:
Cleanup of star setup, core-softening, and MESA file reading modules. Closes issue #91
- setup: Wrap up core-softening code into subroutine `set_softened_core` in the new module `setsoftenedcore`. `set_softened_core` finds the composition, finds the core mass and radius, and calls the softening routines `set_cubic_core` or `set_fixedS_softened_core` depending on the softening option. The cubic core softening module has been renamed from `setsoftenedcore` to `setcubiccore`
- setup: Variable name `hdens` has been changed to `rcore`.
- MESA reading in `rho_profile`: Subroutine `read_mesa` has been upgraded to read both P12-styled profiles and MESA profiles. `read_mesa_file` is no longer necessary.

### DIFF
--- a/src/main/utils_filenames.f90
+++ b/src/main/utils_filenames.f90
@@ -21,7 +21,7 @@ module fileutils
  implicit none
  public :: getnextfilename,numfromfile,basename,get_ncolumns,skip_header
  public :: strip_extension,is_digit,files_are_sequential
- public :: ucase,lcase,make_tags_unique
+ public :: ucase,lcase,make_tags_unique,get_nlines,string_delete
 
  private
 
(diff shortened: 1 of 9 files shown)


src/main/utils_filenames.f90
```diff
@@ -409,4 +430,17 @@ subroutine make_tags_unique(ntags,tags)
 
 end subroutine make_tags_unique
```
danieljprice, line 432: needs a comment here about the purpose of the subroutine

src/setup/density_profiles.f90
```diff
@@ -342,12 +342,12 @@ end subroutine rho_evrard
 !+
 !-----------------------------------------------------------------------
 subroutine read_mesa_file(filepath,ng_max,n,rtab,rhotab,ptab,temperature,&
```
danieljprice, line 344: can we just delete this routine if it is no longer used? It will still appear in the git repo so not truly gone, but would reduce clutter

danieljprice (commented):
just a couple of minor things, otherwise looks good to me

src/setup/density_profiles.f90
themikelau: read_mesa_file is used in moddump_binary, but I will try to replace it with the improved read_mesa.

themikelau: Done

PR #102, opened 2020-12-18
Add possibility to set up an hierarchical triple system embedded in a circum-triple disc

si-mon-jinn:
Add subroutine set_multiple in set_binary.f90 in order to substitute a sink with a binary oriented at will.

### DIFF
--- a/src/setup/set_binary.f90
+++ b/src/setup/set_binary.f90
@@ -25,7 +25,7 @@ module setbinary
 ! :Dependencies: binaryutils
 !
  implicit none
- public :: set_binary,Rochelobe_estimate,L1_point,get_a_from_period
+ public :: set_binary,set_multiple,Rochelobe_estimate,L1_point,get_a_from_period
  public :: get_mean_angmom_vector,get_eccentricity_vector
 
  private
@@ -38,7 +38,12 @@ module setbinary
    ierr_m1   = 1, &
    ierr_m2   = 2, &
    ierr_ecc  = 3, &
-   ierr_semi = 4
+   ierr_semi = 4, &
+   ierr_HIER1 = 5, &
+   ierr_HIER2 = 6, &
+   ierr_subststar = 7, &
+   ierr_Omegasubst = 8, &
+   ierr_missstar = 9
 
 contains
 
@@ -125,7 +130,7 @@ subroutine set_binary(m1,m2,semimajoraxis,eccentricity, &
  endif
  ! exit routine if cannot continue
  if (ierr /= 0) return
-
+ 
  dx = 0.
  dv = 0.
  if (present(posang_ascnode) .and. present(arg_peri) .and. present(incl)) then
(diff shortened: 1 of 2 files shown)


danieljprice:
I think the mpi build failure here is my fault, apologies

src/setup/set_binary.f90
```diff
@@ -254,6 +259,477 @@ subroutine set_binary(m1,m2,semimajoraxis,eccentricity, &
...
+ real                  :: period_ratio,criterion,q_comp,a_comp,e_comp,m_comp,rel_posang_ascnode,rel_arg_peri,rel_incl
+ real                  :: q2,mprimary,msecondary
+ real                  :: alpha_y, beta_y, gamma_y, alpha_z, beta_z, gamma_z, sign_alpha, sign_gamma
+
+ ierr = 0
+ do_verbose = .true.
+ if (present(verbose)) do_verbose = verbose
+
+ 
+  !--- Load/Create HIERARCHY file: xyzmh_ptmass index | hierarchical index | star mass | companion star mass | semi-major axis | eccentricity | period | inclination | argument of pericenter | ascending node longitude
```
danieljprice, line 305: a general comment here is that it would be great to add some documentation about this (just a page in the docs/ directory is all that is required), at least so you at least remember your own workflow here

danieljprice (commented):
see comments below, I am happy to merge but would prefer if you could fix the substantial code overlap between set_binary and set_multiple to avoid having to maintain two copies of the same code

src/setup/set_binary.f90
```diff
@@ -254,6 +259,477 @@ subroutine set_binary(m1,m2,semimajoraxis,eccentricity, &
...
+       theta = f*pi/180.
+       E = atan2(sqrt(1. - ecc**2)*sin(theta),(ecc + cos(theta)))
+    else
+       ! set binary at apastron
+       tperi = 0.5*period ! time since periastron: half period = apastron
+
+       ! Solve Kepler equation for eccentric anomaly
+       call get_E(period,eccentricity,tperi,E)
+    endif
+
```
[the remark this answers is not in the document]
si-mon-jinn: The overlap is due to the fact that set_multiple is written as a generalization of set_binary. If set_multiple is called without specifing a sink to substitute, it behaves exactly as set_binary. 

We kept set_multiple and set_binary separated in order to test the new function without affecting the behaviour of the setup when two sinks are used. The plan is to substitute set_binary with set_multiple eventually. In this way a single function will handle hierarchical systems and binaries, which are the lowest order of a multiple system.

I can fix the code in this way keeping set_multiple as it is and modifying set_binary as a call to set_multiple.

However, if you think is better to keep two separated function for the building of a binary and the building of an hierarchical system I can surely perform the fix you suggested.

Thank you very much for the help and suggestions.

src/setup/set_binary.f90
si-mon-jinn: I will do it soon, thank you

danieljprice:
please can you add a basic documentation page for this so it can be merged?

PR #103, opened 2021-02-03
Star

themikelau:


danieljprice:
this pull request is ok, but the bug with compiling in single precision needs to be fixed in order for it to be merged

themikelau:
> this pull request is ok, but the bug with compiling in single precision needs to be fixed in order for it to be merged

Will look at that!

PR #105, opened 2021-02-09
Fixes for h2cooling

rieder:
Discussed with @jameswurster

### DIFF
--- a/src/main/force.F90
+++ b/src/main/force.F90
@@ -2872,7 +2872,7 @@ subroutine finish_cell_and_store_results(icall,cell,fxyzu,xyzh,vxyzu,poten,dt,dv
 
        ! cooling timestep dt < fac*u/(du/dt)
        ! Note: Why is this not used for *all* energy changes?  Regrettably, Sedov will crash if this timestep is included since eni0 = 0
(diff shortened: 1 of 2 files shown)


danieljprice:
happy to look at this, but please fix the build failures on this PR

jameswurster:
The issue with this commit is the timestep constraint.  Steven and I have discussed it.  

This has only become an issue once cooling was moved to step.  The heating here still contributes to dudt, so arguably there should be a corresponding timestep.  Indeed, Steven found that this was necessary for Amuse when using the h2cooling (done in Step).  However, as we can see, tests fail when heating is here and cooling is elsewhere.  In this case, we fail since there are particles with eni = 0, meaning dtcool = 0.  If we add the constraint if (eni > epsilon(eni)), then I found it will still fail since as soon as eni > epsilon, we will get a dtcool << 1, and we will stop due to the timestep being too small.  

As a simple fix, I suggest putting a pre-processor flag around the if-statement so that it is always on if amuse is being used or if cooling_explicit = .true.  For the long term, since heating is just an `anti-cooling' process, I wonder if it all should be moved to step to be implicitly calculated and where all the dudt terms are self-consistently calculated on particles on the same step and at the same temperature.

PR #110, opened 2021-03-01
Resolve the code overlap between set_binary and set_multiple

si-mon-jinn:
set_binary is now treated as a lower level routine called by set_multiple in order to build hierarchical systems.


danieljprice:
thank you Simone - much improved - documentation is also great here!

PR #113, opened 2021-03-10
Fixing missing i and . (got_orig = .true.)

distamio:


distamio:
accidental

dmentipl:
I think this is correct. Perhaps it should be merged.

PR #117, opened 2021-03-23
fix unit conversion of distance and mass in moddump dustadd.f90

crislong:
In the file "moddump_dustaff.f90", when the 2 fluids method is selected, the code takes the dimension and the density of dust grains in physical units (cm, g/cm^3) and does not convert them in code units. I fixed this problem by converting the size and the density in code units.

Cristiano Longarini
cristiano.longarini@unimi.it


crislong:
Good morning professor Price, I am Cristiano Longarini, a PhD student at the University of Milano under the supervision of Giuseppe Lodato. 
Using the moddump dustadd.f90, I noted that the dimension and the density of the added dust particle is wrong (a problem of physical-code units). I tried to fix it, I hope everything is fine!
Looking forward to knowing what you think about it, 
Cheers

Cristiano

danieljprice:
looks great, thanks for the pull request

PR #122, opened 2021-05-14
New sink formation criteria

jameswurster:
added icreate_sinks=2, where the checks for sink formation are performed on the particles within 2h of the candidate gas particle; this is what sphNG does.  This method facilitates the quicker formation of sink particles than the original/default icreate_sinks=1 which checks all particles within h_acc.

### DIFF
--- a/src/main/initial.F90
+++ b/src/main/initial.F90
@@ -521,6 +521,11 @@ subroutine startrun(infile,logfile,evfile,dumpfile)
  endif
  call init_ptmass(nptmass,logfile,dumpfile)
  if (gravity .and. icreate_sinks > 0) then
+    if (icreate_sinks==1) then
+       write(iprint,*) 'Sink will be created if particles within h_acc meet formation criteria.'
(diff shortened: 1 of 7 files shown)


danieljprice:
could we just make icreate_sinks=2 the default?

jameswurster:
I left icreate_sinks=1 for legacy (although I'm not sure how many people using Phantom use sink creation).  I can happily remove it if you prefer.

jameswurster:
@danieljprice I have revised the sink formation to always check the particles within 2h; this can be simply reverted by modifying two hard-coded lines in ptmass.f90

The bug in set_unifdis was accidentally implemented by me when I was fixing another bug.  This only became apparent when running make test rather than make testkd, which is the pipeline's tests.

When the Sedov test is run with global timestepping (as in make test), the timesteps are slight longer than with individual timestepping, thus the energy error is slightly higher; the tolerance has been increased to permit the test to pass when using global timestepping.

PR #126, opened 2021-06-11
Thermal Floor + added Balsara & Kim defaults

jameswurster:
I have added an option to impose a thermal floor to prevent a simulation from crashing with utherm < 0.  This is controlled by the logical use_Tfloor and the value Tfloor in eos.F90.  This should be used only as a last resort, which is why they're hardcoded and not in the .in file.  By default, the thermal floor is off.

@rieder and myself have found a few extreme simulations where this is necessary, and simply reducing the timestep does not work as a result of the predictor-corrector method.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -746,6 +746,7 @@ endif
 
 ifeq ($(SETUP), balsarakim)
 #   Balsara-Kim 2004
+#   set BalsaraKim=.true. in setup_unifdis.f90 to initialise the correct defaults
     PERIODIC=yes
     SETUPFILE= setup_unifdis.f90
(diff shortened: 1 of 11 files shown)


jameswurster:
The MPI build failure is a github issue and not a result of these modifications.

jameswurster:
The above commits are to add the defaults for the Balsara & Kim (2004) test.  By default, these defaults are off, but they can be turned back on by setting balsarakim = .true. in setup_unifdis.  
I have also modified partinject to react properly to change the timestep.   Previously, global timestepping and individual timestepping when particles are modified by not updated were not reacting fast enough.

PR #129, opened 2021-07-09
 (moddump_binary) add option to set up binary by adding a star from another dump file

themikelau:


themikelau:
Works when second dump file is same as the first dump file, but facing issues when they are different—possible errors arising from different particle masses?

danieljprice:
can you elaborate on what the issues are?

themikelau:
I am getting a bunch of "WARNING: particles of same type at same position: ". But this is probably just a symptom.

PR #135, opened 2021-07-22
(ptmass) use logfile prefix for sink ev prefix instead of dumpfile pr…

### LINKED ISSUE #134: Incorrect sink ev file naming
Illustrative example:

With the following options in the infile:
```
logfile =  L3/dumps/L301.log   ! file to which output is directed
dumpfile =  L3/dumps/dump_00100   ! dump file to start from
```
When running Phantom v2021.0.0, the sink evfiles are named in the form of, e.g., `dumpSink0001N.ev`, instead of, e.g., `dumpSink0001N01.ev` for the first file and etc. This resulted in data loss where the subsequent runs replace `dumpSink0001N.ev` instead of creating `dumpSink0001N02.ev` and `dumpSink0001N03.ev`, etc.

However, the log files and ev files are named correctly, i.e., `L3/dumps/L301.log, L3/dumps/L302.log, ...` and `L3/dumps/L301.ev, L3/dumps/L302.ev, ...`

themikelau:
…efix

Would fix #134 . However, need to check that getting the ptmass ev file prefix with `logfile(1:idot-3)` isn't dangerous—this change assumes the logfile number takes up two characters.

### DIFF
--- a/src/main/ptmass.F90
+++ b/src/main/ptmass.F90
@@ -1366,7 +1366,7 @@ subroutine init_ptmass(nptmass,logfile)
  !
  idot = index(logfile,'.')
  if (idot==0) idot = len_trim(logfile) + 1
- write(pt_prefix,"(a)") logfile(1:idot-3)
+ pt_prefix = logfile(1:idot-3)
 
  !
  !--Define file name components and finalise suffix & open files
--- a/src/main/ptmass.F90
+++ b/src/main/ptmass.F90
@@ -36,7 +36,7 @@ module ptmass
 !
  use dim,  only:maxptmass
  use part, only:nsinkproperties,gravity,is_accretable
- use io,   only:iscfile,iskfile
+ use io,   only:iscfile,iskfile,id,master
  implicit none
  character(len=80), parameter, public :: &  ! module version
     modid="$Id$"
@@ -1361,6 +1361,8 @@ subroutine init_ptmass(nptmass,logfile)
  character(len=*), intent(in) :: logfile
  integer                      :: i,idot
  character(len=150)           :: filename
+
+ if (id /= master) return ! only do this on master thread
  !
  !--Extract prefix & suffix
  !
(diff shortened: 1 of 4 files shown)


danieljprice:
any idea why indentation changes suddenly kicked in on libphantom-evolve.F90?

src/main/ptmass.F90
```diff
@@ -1356,18 +1356,18 @@ end subroutine ptmass_create
...
+ integer                      :: i,idot
  character(len=150)           :: filename
  !
  !--Extract prefix & suffix
  !
- idash = index(dumpfile,'_')
- write(pt_prefix,"(a)") dumpfile(1:idash-1)
  idot = index(logfile,'.')
  if (idot==0) idot = len_trim(logfile) + 1
+ write(pt_prefix,"(a)") logfile(1:idot-3)
```
danieljprice, line 1369: it doesn't make sense to use a write statement here to write a string to a string (even though this was done in previous code). Should just say pt_prefix = logfile(1:idot-3)

src/utils/libphantom-evolve.F90
```diff
@@ -263,162 +263,162 @@ subroutine finalize_step(infile, logfile, evfile, dumpfile)
      .or.(time >= tmax)            &
      .or.((mod(nsteps,nout)==0).and.(nout > 0)) &
      .or.(nsteps >= nmax)) then
- !--modify evfile and logfile names with new number
- if (noutput==1) then
-    evfile = getnextfilename(evfile)
-    logfile = getnextfilename(logfile)
- endif
- dumpfile = getnextfilename(dumpfile)
```
danieljprice, line 271: am puzzled why the indentation got changed here but not in previous pull-requests. Seems ok though

src/main/ptmass.F90
themikelau: Okay

danieljprice:
I think the MPI failure might be a genuine problem here, e.g. maybe logfile is only defined on the master thread? or both threads try to open the Sink.ev file where only one should?

danieljprice:
I am still hunting what is causing the MPI failure here - seems to work fine on my laptop, possibly due to newer compiler version in the pipelines, or masking some other problem

PR #167, opened 2021-08-31
Bug fixed for u<0

Fitz-Hu:
Bug fixed for u<0
Add star density profile option in grtde
Putting star on parabolic orbit in moddump_tidal


jameswurster:
Please check to ensure that your change to the cooling timestep does not have unintended consequences, such as making the Sedov test impossibly slow (as noted in the comment you deleted).  I recently added an optional thermal floor as a work-around to prevent u < 0 since timestep modifications (including the one you committed) was not sufficient for rapidly evolving regions.

Please also be cautious when making a change that will affect many groups, such as changing the default units in a setup file.  Since this setup is widely used, it would be safer to prompt the choice of units rather than force GR units on everyone.

src/setup/setup_star.f90
```diff
@@ -239,8 +239,8 @@ subroutine setpart(id,npart,npartoftype,xyzh,massoftype,vxyzu,polyk,gamma,hfact,
  !
  ! set units
  !
- call set_units(dist=udist,mass=umass,G=1.d0)
- !
+ !call set_units(dist=udist,mass=umass,G=1.d0)
+ call set_units(mass=umass, c=1.d0, G=1.d0)
  ! set up particles
```
danieljprice, line 244: see comment by James -- we should not make geometric units the default here!!

danieljprice: probably best just to leave this line commented out so the default units are as previously

src/utils/moddump_tidal.f90
```diff
@@ -263,7 +282,11 @@ subroutine read_setupfile(filename,ierr)
  call read_inopt(theta, 'theta', db,min=0.,errcount=nerr)
  call read_inopt(phi,   'phi',   db,min=0.,errcount=nerr)
  call read_inopt(r0,    'r0',    db,min=0.,errcount=nerr)
+<<<<<<< HEAD
+ call read_inopt(ecc,   'ecc',   db,min=0.,max=1.,errcount=nerr)
+=======
```
danieljprice, line 287: this is the cause of the build failure, just some of the git conflicts caused by code merge
i.e. get rid of the <<<< HEAD and ===== and just leave the line that you want here

src/utils/moddump_tidal.f90
```diff
@@ -263,7 +282,11 @@ subroutine read_setupfile(filename,ierr)
...
  call read_inopt(ecc,   'ecc',   db,min=0.,errcount=nerr)
+>>>>>>> bbc141093e935c089608d0c98943c71dd8a7a50b
```
danieljprice, line 290: same with this one, just delete the >>>> here

danieljprice (changes requested):
@jameswurster including the cooling timestep does indeed solve the problem we were having with u < 0 in rapidly expanding regions, and we don't want to enforce a temperature floor for this case. We added a condition so that the sedov test should still run ok, i.e. we tests for u > tiny so as not to end up with dtcool= 0

jameswurster:
@danieljprice thanks for the clarification regarding the timestep.  My previous tests were detrimental to Sedov (hence the comment), but I think I was using epsilon rather than tiny for both f & u.

danieljprice (approved):

PR #172, opened 2021-09-13
Sinks can now merge

jameswurster:
Added algorithm to permit sink mergers.  They unconditionally merge if they pass with r_merge_uncond of one another, and will merge if they pass within r_merge_cond and they are bound.  By default, both radii are 0 to prevent mergers.  Once a particle as merged, its mass is set to negative; this has been accounted for throughout the code.

### DIFF
--- a/src/main/centreofmass.f90
+++ b/src/main/centreofmass.f90
@@ -130,13 +130,15 @@ subroutine get_centreofmass(xcom,vcom,npart,xyzh,vxyzu,nptmass,xyzmh_ptmass,vxyz
  if (id==master .and. present(xyzmh_ptmass) .and. present(vxyz_ptmass) .and. present(nptmass)) then
     do i=1,nptmass
        pmassi = xyzmh_ptmass(4,i)
-       totmass = totmass + pmassi
-       xpos  = xpos  + pmassi*xyzmh_ptmass(1,i)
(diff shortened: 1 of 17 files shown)


jameswurster:
A corresponding commit to Splash has been made to prevent plotting the merged sink that is tagged with negative mass.

jameswurster:
Note: test failure options seem to be a result of a failure in the test bot not having the proper SSH keys, not a result of my updates.

danieljprice:
it's a little hard to decipher from the logs, but the failures here are in:

star
radstar

jameswurster:
Thanks for letting me know where the bugs were.  These logs are harder to decipher than the bitbucket logs, especially when some faults are the developers (e.g. mine) and some seem to be the system not running or having a permission... 
Hopefully the above patch will be the last of my bugs.

jameswurster:
The 
mpi / MPI DEBUG=no OPENMP=yes (pull_request) 
test passes on my local mac and on my fork of Phantom.  This failure is from the dE/dt being too large that we've been seeing sporadically for a while now.

danieljprice:
I reran the test suite and this passed. I guess we should alter the tolerance slightly on the dE/dt test.

Thanks for the contributions!

danieljprice:
btw the benchmark failures were because I needed to renew the project running the data server, so it could not download the required files to run the benchmarks...

PR #176, opened 2021-10-07
(star formation) added star formation setup and bug fix with Bonnor e…

raomoomoo:
…bert sphere boundary condition

### DIFF
--- a/src/setup/setup_disc.f90
+++ b/src/setup/setup_disc.f90
@@ -25,21 +25,29 @@ module setup
 !   - Ratm_in       : *inner atmosphere radius (planet radii)*
 !   - Ratm_out      : *outer atmosphere radius (planet radii)*
 !   - accr1         : *single star accretion radius*
+!   - accr1a        : *tight binary 1 primary accretion radius*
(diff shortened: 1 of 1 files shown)


danieljprice:
but the test failure here was as you predicted @jameswurster from Sedov test going wrong with the recent merge of the du/dt timestep condition to try to prevent the u < 0 failure

jameswurster:
I still don't understand what the bug is that you fixed...
i) the sound speed in the medium is currently set to ensure pressure equilibrium (assuming a barotropic EOS); the updated version has a higher background pressure, such that the outer regions of the BE sphere have less gas pressure than the background, meaning that the background will promote collapse rather than hinder expansion.
ii) the original rho0 was 2x the background density, so we did account for fluctuations in density without an immediate temperature plunge.  In my personal working version, I'm using 4x (assuming a density contrast of 30, although I suspect the exact factor does not matter as long as it's > 1).  By default, the barotropic EOS includes a smooth 0.5dex transition across rho0 to prevent a step-function change in temperature.  

Clarification on your changes would be much appreciated.  Thanks

PR #185, opened 2021-11-11
Wind updates

lsiess:
- fix several bugs in wind calculation
- implement variable mean molecular weight  in EOS (needed because of H2 formation in the wind)
- implement molecular wind cooling (not tested and limited to binary systems - Ward's student developments)
- add python script to read phantom dumps


danieljprice:
there's a conflict here because we already made the change for the variable mu in the equation of state routine

PR #186, opened 2021-11-15
Gravitational wave emission

danieljprice:
This is a pull request from Martina Toscani based on the paper https://ui.adsabs.harvard.edu/abs/2021arXiv211105145T/abstract

The functionality is as follows:
- can compute and output gravitational waveforms from any phantom simulation, using the quadrupole approximation
- waveforms are written to the .ev file
- currently the option to do this is switched on automatically if code units with G=c=1 are employed (in either the Newtonian or GR codes)
- works with both gas particles and sink particles, but currently assumes all SPH particles have the same mass (so would not work with additional particle types like dust/stars/etc). Would be a minor change to amend this but no obvious use case
- in principle would work to compute gravitational waves from any simulation (e.g. star formation), just will be very low amplitude!
- can be done either during the simulation, or in post-processing using the analysis_gws.f90 subroutine

I've fixed most of the compile time and formatting issues. A few structural issues need fixing before merge, namely:

- calc_strain currently does two things: calculates the strain AND writes several files. These should be split into two different functions (one to compute the strain, the other to write the file)
- with the above, can reduce code overlap further in analysis_gws.f90 and utils_gravwaves.f90
- some rotation is implemented in the calculate_strain routine, but relies on reading parameters from a grtde.setup file which may or may not exist. These parameters should be put in the .in file (and propagated from the .setup to the .in file if necessary)
- cleanup some of the remaining commented-out code
- should add some basic tests to the automated test suite, e.g. comparison with exact solution for gravitational wave emission from binary


danieljprice:
Updates so far implement several changes:
- added test for gravitational wave emission from circular binary in test_ptmass.f90
- fixed bug with MPI code, gravitational wave emission now correct with MPI also
- split test_ptmass into smaller subroutines for modularity

Remaining things:
- need to remove dependence on grtde.setup
- could just add gravitational wave computation (and number of viewing angles?) as an option in the .in file

danieljprice:
this should now be ready to merge
- removed explicit read/write of grtde.setup, but same functionality as the rotation angle is now automatically set in the input file from setup_grtde.f90
- rotation angles are now options in the .in file
- gravitational wave emission is now a T/F option in the .in file

PR #187, opened 2021-11-15
(testeos) add forward-backward tests for MESA and idealplusrad EoSs

themikelau:


danieljprice:
```
Well, my terminal's locked up, and I ain't got any Mail,
	And I can't recall the last time that my program didn't fail;
I've got stacks in my structs, I've got arrays in my queues,
	I've got the : Segmentation violation -- Core dumped blues.

If you think that it's nice that you get what you C,
	Then go : illogical statement with your whole family,
'Cause the Supreme Court ain't the only place with : Bus error views.
	I've got the : Segmentation violation -- Core dumped blues.

On a PDP-11, life should be a breeze,
	But with VAXen in the house even magnetic tapes would freeze.
Now you might think that unlike VAXen I'd know who I abuse,
	I've got the : Segmentation violation -- Core dumped blues.
		-- Core Dumped Blues
```
courtesy of the "fortune" command 😄

PR #188, opened 2021-11-22
Add ideal gas + radiation + recombination EoS (from HORMONE, Hirai et al. 2016)

themikelau:
- Add `ionization.f90` containing module `ionization_mod` taken directly from HORMONE (might not compile in single precision), tweaked to use constants from `physcon`.
- Added new EoS (ieos=20) and new module `eos_gasradrec`.
- Allow equation of state to take local mean molecular weight (mu), hydrogen fraction (X), and metallicity (Z) rather than from module global variables `gmw`, `X_in`, and `Z_in`. `mu` is outputted from the `equationofstate` subroutine if using the gas+rad+recombination EoS.
- Store mu, X, Z in `eos_vars` array.
- Add forward-backwards calculation tests for the gas+rad EoS and gas+rad+recombination EoS.
- Set up star with variable composition.


src/setup/setup_star.f90
```diff
@@ -718,20 +722,17 @@ subroutine write_setupfile(filename,gamma,polyk)
  case(2)
     call write_inopt(gamma,'gamma','Adiabatic index',iunit)
     if (input_polyk) call write_inopt(polyk,'polyk','polytropic constant (cs^2 if isothermal)',iunit)
-    if ((isoftcore <= 0) .and. (.not. use_variable_composition)) call write_inopt(gmw,'mu','mean molecular weight',iunit)
+    if (.not. use_variable_composition) call write_inopt(gmw,'mu','mean molecular weight',iunit)
```
themikelau, line 725: Note to self: Restore this line to original version.

src/setup/setup_star.f90
```diff
@@ -837,15 +838,18 @@ subroutine read_setupfile(filename,gamma,polyk,ierr)
...
     ! if softening stellar core, composition is automatically determined at R/2
-    if ((isoftcore <= 0) .and. (.not. use_variable_composition)) then
+    if ( (.not. use_variable_composition) .and. (isoftcore <= 0)) then
        call read_inopt(X_in,'X',db,errcount=nerr)
        call read_inopt(Z_in,'Z',db,errcount=nerr)
     endif
  case(12)
     ! if softening stellar core, mu is automatically determined at R/2
-    if ((isoftcore <= 0) .and. (.not. use_variable_composition)) call read_inopt(gmw,'mu',db,errcount=nerr)
+    if ( (.not. use_variable_composition) .and. (isoftcore <= 0)) then
```
themikelau, line 850: Note to self: Restore to original version

ryosuke-hirai (commented):
All the hard-wired double precision numbers in `eos_gasradrec.f90` and `ionization.f90` should be rewritten as single precision numbers (e.g. `1d0` -> `1.`), to make it pass the single precision compilation test.
Also, the reference to the gasradrec EoS should be Hirai+20 (Appendix C), not Hirai+16.

themikelau:
> All the hard-wired double precision numbers in `eos_gasradrec.f90` and `ionization.f90` should be rewritten as single precision numbers (e.g. `1d0` -> `1.`), to make it pass the single precision compilation test. Also, the reference to the gasradrec EoS should be Hirai+20 (Appendix C), not Hirai+16.

Thanks! Will fix this.

ryosuke-hirai (commented):
There's a couple of double precision's left in `eos_gasradrec.f90` around line 37-64.

danieljprice:
yep, build failures are still in SETUP=blob, you can just try compiling with "make SETUP=blob" on your own machine to reproduce these

danieljprice (approved):
all seems ok to me, would suggest just to run the bots on your branch to fix minor formatting issues before merging.
i.e. 
```
cd phantom/scripts
./bots.sh --commit
```

Issue #194, opened 2021-12-06
Make cleanall the default


ttricco:
"Make clean" removes the compiled code
"Make cleanall" removes the compiled code + binaries

Change to have "make clean" perform "make cleanall" and remove "make cleanall". But don't remove current working directory copies.

conradtchan:
Implemented in #196

PR #196, opened 2021-12-06
minor cleanup, including resolving make clean (issue #194)

jameswurster:


src/setup/setup_cluster.f90
```diff
@@ -78,6 +79,9 @@ subroutine setpart(id,npart,npartoftype,xyzh,massoftype,vxyzu,polyk,gamma,hfact,
  logical                      :: inexists,setexists
  logical                      :: BBB03 = .true. ! use the BB03 defaults, else that of a YMC (S. Jaffa)
 
+ !--Ensure this is pure hydro
+ if (mhd) call fatal('setup_cluster','This setup is not consistent with MHD.')
```
danieljprice, line 83: can you explain the reason for this? The cluster setup at the moment would be a re-do of Price & Bate (2008) if run with MHD. Is there a reason why we should forbid this?

jameswurster: I was under the impression that magnetic fields do not behave at open boundaries when evolved explicitly, hence the shift to the sphere-in-box setup.  I think the magnetic fields in Price & Bate (2008) behaved since they were evolved using vector potentials.

PR #200, opened 2021-12-21
Radiation star setup

themikelau:


src/main/eos.F90
```diff
@@ -709,6 +712,10 @@ subroutine read_options_eos(name,valstring,imatch,igotall,ierr)
     read(valstring,*,iostat=ierr) gmw
     ! not compulsory to read in
     if (gmw <= 0.)  call fatal(label,'mu <= 0')
+ case('gamma')
```
danieljprice, line 715: this one is a bit dangerous. The idea is that gamma should be set from the initial conditions and not changed after that. So it is currently written/read to/from the header of the dump files. Is there a particular use case that requires gamma to be a parameter in the .in file? It can happily be a parameter in the .setup file but for consistency I think it is better to leave it in the dump header.

danieljprice, line 715: this one is a bit dangerous. The idea is that gamma should be set from the initial conditions and not changed after that. So it is currently written/read to/from the header of the dump files. Is there a particular use case that requires gamma to be a parameter in the .in file? It can happily be a parameter in the .setup file but for consistency I think it is better to leave it in the dump header.

themikelau: Okay I am happy to delete this—I just forgot that gamma is read from the dump header.

PR #216, opened 2022-02-03
Sink heating

themikelau:
Type of PR: 
Implement ability for sink particles to heat gas particles within softening radius.

Description:
* Luminosity of point mass is specified via the `ilum` column. A new column, `imass_enclosed`, has been added to store the mass enclosed within twice the softening length, 2*hsoft.

Testing:
* Observed reasonable stability of star with a sink-particle core radiating at the stellar luminosity, with radiation transport turned on.

Did you run the bots?
No.


danieljprice:
@themikelau this should be ready to go. I cleaned up a couple of things in test_eos so please check everything is ok from your side (in particular I made a helper routine to avoid the repeated code to setup the rho, T grid; since the choice seemed arbitrary I made the limits the same for both, unless there is a reason you wanted different limits tested for idealplusrad vs. hormone?)

src/utils/moddump_sinknuclearburning.f90
```diff
@@ -0,0 +1,41 @@
...
+ use part,           only:xyzmh_ptmass,vxyz_ptmass,nptmass,ilum
+ use ptmass_heating, only:Lnuc
+ use units,          only:unit_energ,utime
+ integer, intent(inout) :: npart,npartoftype(:)
+ real,    intent(inout) :: massoftype(:),xyzh(:,:),vxyzu(:,:)
+ real                   :: Lnuc_cgs
+ integer                :: core_sink_id
+
+ Lnuc_cgs = 1.672e38
+ Lnuc = Lnuc_cgs / unit_energ * utime
```
danieljprice, line 33: might be worth defining a unit_luminosity in set_units_extra in units.f90 to simplify this expression?

themikelau:
@danieljprice Thanks for cleaning up test_eos. Your changes are fine if the test passes.

PR #218, opened 2022-02-04
Dust arrays

jameswurster:
Type of PR: 
Bug fix

Description:
Changed array operations from acting on the entire arrays to just on the filled elements. 

Testing:
Does not segfault.  Prior to the fix, there were issues when the arrays were not the same size.

Did you run the bots? yes (hence the commit of part.F90)


jameswurster:
Several dust properties printed to the .in file require information from the dump file.  This is acceptable for the actual .in files, however, the .in information that is printed to the header is printed before the dump file is read.  Therefore, default (and likely incorrect) values were being printed.  This commit prevents any information from being printed to the header that requires knowledge from the dumpfile.  A warning is included to let the user know the list is incomplete.

PR #225, opened 2022-02-10
Cooling, timing, & misc

jameswurster:
Type of PR: 
Multiple

Description:
Bug fix: Corrected explicit Koyama&Inutuska cooling
New Physics: added implicit Koyama&Inutuska cooling
Updated numerics: set_ellipse can now use random particle placement
Updated numerics: set_unifdis can use particle number rather than psep
Updated numerics: if dt < dtmax*0.5**maxbins, the exit will now occur in step after printing useful diagnostic info.
Bug fix/update: updated how dtmax can increase if dtmax is permitted to change with wall time

Testing:
These changes have been successfully used in test & production runs.

Did you run the bots? yes

(Sorry for the multiple topics in a single PR; each topic is localised to a single commit)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,28 +29,28 @@ Thomas Reichardt <thomas.reichardt@students.mq.edu.au>
 Simon Glover <glover@uni-heidelberg.de>
 Martina Toscani <mtoscani94@gmail.com>
 Jean-François Gonzalez <Jean-Francois.Gonzalez@ens-lyon.fr>
-Benedetta Veronesi <benedetta.veronesi@unimi.it>
 Simone Ceppi <simo.ceppi@gmail.com>
+Benedetta Veronesi <benedetta.veronesi@unimi.it>
 Christopher Russell <crussell@udel.edu>
-Alessia Franchini <alessia.franchini@unlv.edu>
 Megha Sharma <msha0023@student.monash.edu>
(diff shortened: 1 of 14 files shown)


danieljprice:
failing with SYSTEM=ifort on SETUP=
nshwdisc
balsarakim
test

jameswurster:
Thanks @danieljprice for letting me know about which ones are failing.  I am having difficulty reproducing the errors, and not sure if they are my errors or not.  In the log, the error for balsarakim & test is
ifort: error #10052: could not checkout FLEXlm license

and the error for nshwdisc is
nifort: command line error: option '-openmp' is not supported. Please use the replacement option '-qopenmp'

I'm not sure what these errors are, or if it's an error with the host machine. 

For a tricky error I may have found, when I ran test using ifort and DEBUG=no, then I was getting a failure that the header has 8 errors instead of zero.  Running DEBUG=yes with ifort or gfortran with both debug options yields no errors.  In my commit I added a variable to the header, but I don't see how that could cause the error in just one case.

I found a trivial error in a print statement, so I've commit that change, and let it trigger a re-running of the tests to see if these errors are reproducible on the host.

danieljprice:
looks like it was errors on the host machine, re-triggering the bots was the right thing to do!

PR #250, opened 2022-03-02
Working gas+rad primitive variable solver with gr

Fitz-Hu:
Type of PR: 
Bug fix / new physics

Description:
Implemented bisection method in gr primitive variable solver for gas+rad

Testing:
Tested in testgr

Did you run the bots? no


danieljprice:
this is working but the TDE benchmark needs updating because the timestep has changed slightly

Issue #87, opened 2020-11-10
Remove store_temperature option


danieljprice:
the store_temperature option is now redundant (temperature is always stored) and can be removed from the code. 

Since the equation of state also now returns the temperature for all choices of ieos, we should also remove all superfluous "get_temperature" routines.

Suggest that temperature should only be written to the dump file if a non-ideal equation of state is used, to avoid wasting disk space

_Originally posted by @danieljprice in https://github.com/danieljprice/phantom/pull/86#issuecomment-724400949_

PR #252, opened 2022-03-02
Temperature & Pressure i/o

jameswurster:
Type of PR: 
modification to existing code

Description:
-min/max/ave temperature is now printed to the .ev file
-pressure is read in from the dump file if present (useful for phantomanalysis)
-remove unused variables, including maxtemp

Testing:
Ran test simulation; ran test suite

Did you run the bots? 
yes, but only committed the changes to the files I modified
(Will we run into issues if we multiple pull requests contain the same bot-changes?)

QUESTION:
Is store_temperature and/or eos_is_non_ideal still necessary?  These seem to be related mostly to whether or not temperature is printed to the dump file; even if we want conditional printing to the dumpfile, the former appears obsolete.  It might be easiest to remove these and always print temperature & pressure to the dump file for ease of Splashing & phantomanalysis.  
The caveat is that these two variable control how denij is calculated in force.F90.  I am unsure how to proceed with that calculation.


danieljprice:
See #87 

Only remaining issue is that writing of temperature to either the dump file or .ev file needs be conditional in some way. For example, it only makes sense to report temperature if physical units are being employed. Disk space is also one of our major constraints when running phantom, so we should only dump temperature when absolutely necessary.

danieljprice:
in other words we could replace if (store_temperature) in readwrite_dumps with some more sensible logic.

jameswurster:
Thanks for pointing out the open issue.  I have no problem cleaning it once we agree on how (coding will be a nice break from all the reading & writing I've been doing lately).  It looks like a lot of the depreciated code has already been stripped out.  

To save memory in the dump file, we can use the eos_is_non_ideal logical when deciding to print temperature.  The contents will need to be updated whenever a new ieos is added, though, but that's not too dissimilar to currently needing to select the correct flags.  

Regarding temperature in the .ev file, I added this since it would have made my student's life easier.  We wanted to track T_max vs time with a higher cadence than the dump file.  Since we're using ieos=8, he reconstructed T_max using rho_max, but it would have made life easier if T_max was readily available.  If you want, we can conditionally print it to the .ev file if the conditions are such that we are printing temperature or pressure to the dump file.

jameswurster:
Type of PR:
Bug fix

Description:
-removed the incorrect usage of ufloor in eos when ieos=2; (bug added on commit f0dffbf)
-re+added call fatal when eni < 0 (call was removed on commit 778139b)

Testing:
ran test suite

Did you run the bots?
no

danieljprice:
have merged for now as ok as-is, we can just open a new p-r for the code to get rid of store_temperature altogether

jameswurster:
Much appreciated.
I'll start removing the store_temperature this week (hopefully).

PR #263, opened 2022-03-10
eos & store_temperature

jameswurster:
Type of PR: 
modification to existing code 

Description:
-Removed logical store_temperature and it's associated pre-processor statement (issue #87) since temperature is always stored.  
-temperature is now a non-optional variable to be passed to/from equationofstate
-re-tasked get_temperature & get_spsound to be wrapper routines to remove duplicated code; added get_pressure
-removed unused subroutines/functions in eos
-reordered eos for readability
-general cleanup regarding eos-related called
-Note: there was a major issue when I updated from the repo due to us both modifying Makefile; the version here should be the simplified one without the store_temperature pre-processor commands.  It may look like I re-wrote the Makefile, but I did not.  

Testing:
ran testsuite & a few toy simulations

Did you run the bots? yes


jameswurster:
This PR is failing on the tde benchmark test.  This is the same test that is failing for several of Conrad's PR's, so is this an issue that was already committed to the repo?  I am happy to bug-hunt if I caused the issue with my modifications.

danieljprice:
the failure is not caused by you (it's due to the benchmarks needing an update), however I merged some of the other PRs so there are some conflicts to be resolved here

the only part I think could simplify matters here is that temp should still be an optional argument to equationofstate. This would avoid having to modify all of the analysis routines that call eos. It's also weird in things like shock tubes to be get temperatures when no physical units are set.

jameswurster:
I can revert to an older commit and re-add the optional temperature if you want.  However, for cleanliness and simplicity, I do not think temperature should be optional.  Whenever equationofstate is called from the main code, temperature is passed, and I think we should clean the code based upon Phantom and not Phantomanalysis.  Moreover, we should likely be using get_temperature, get_spsound, get_pressure in the analysis files anyway.

The simplification I was wondering about would be to replace 
get_temperature(ieos,...,
get_spsound(ieos,...,
get_pressure(ieos,...,
with

get_TempCsPres('temp',ieos,...

danieljprice:
I'm ok with leaving temperature non-optional.  There are only two conflicts here, I don't think it's catastrophic. I'm happy enough with your get_pressure, get_spsound routines, I don't think they need to be combined further

jameswurster:
I updated my version of Phantom and resolved all conflicts locally so that I could test before recommitting.  I've successfully done this on my branch, but nothing has been pulled across here.  Did I have to use the above button to resolve these issue via a web interface?  

I don't know how to resolve this issue now and get my merg-able files across.

jameswurster:
the bots commit got pulled across, but I thought there should have been more changes.  
I'm really not sure what is going on with this PR after these merge conflicts.  

I made the mistake of reordering the subroutines in eos.F90 into a useful and logical order that better matched the other subroutines in Phantom; since a few students modified these subroutines simultaneously, that caused the merge issue.

PR #282, opened 2022-04-11
(eos.F90) Fixed bug in Eos 14: the original prescrition was commented…

eragusa:
… and replaced by one normalised by the mass of the primary instead of using the total binary mass.

Type of PR: 
Bug fix 

Description:
Restored to the original Farris et al. (2014) prescription implemented in Ragusa et. al (2016).

Testing:
Compared the results with other codes

Did you run the bots? yes


danieljprice:
We need a unit test for this eos to prevent issues like this in future

PR #298, opened 2022-06-06
./phantomsetup --maxp=10000000 instead of recompiling with MAXP=

danieljprice:
Type of PR: 
other

Description:

Phantomsetup now accepts the --maxp flag which controls the memory allocation. This is a simple workaround for the fact that we have not fully integrated dynamic memory allocation into the individual setup routines called by phantomsetup. The memory allocation is semi-static in the sense that it is done before the setup routines are called. The new flag enables this to be a runtime option which avoids the need to recompile phantomsetup.

phantom itself has not used or required any MAXP flag for quite a while, memory is dynamically allocated upon reading the dump file. Ideally the MAXP flag should be deleted from all SETUPs and should no longer have any effect. We could now also remove maxp_hard in principle.

Testing:
```
~/phantom/scripts/writemake.sh disc > Makefile
make setup
./phantomsetup --maxp=10000000 disc
```

Did you run the bots? no


jameswurster:
Unfortunately, we cannot remove maxp_hard since large, static arrays are required in a few cases, namely particle injection, particle splitting, and dynamic boundaries (for the latter, see Wurster & Bonnell in prep)

danieljprice:
these cases can also be solved by giving --maxp as a command line flag

jameswurster:
I see what you're doing.  I agree it would be nice to remove the compile-time parameter.  However, if we remove maxp_hard, then will the above cases be forced to include a --maxp command line option when running?  This seems like something we'd want to avoid....

danieljprice:
good point, seems like maxp_hard might have some life in it yet :) Might be good discussion at next dev telecon to see if we can think of a better way

PR #303, opened 2022-06-26
Bonnor-Ebert sphere debugged (again)

jameswurster:
Type of PR: 
Bug fix

Description:
Option 6 of the BE sphere (inputting physical mass & radius) is now fixed

Testing:
Ran all 6 input options describing the same sphere and verified all inputs produced the same sphere.

Did you run the bots? no


danieljprice:
This sounds like a great thing to automatically check in the test suite...

PR #308, opened 2022-07-22
Stratified EOS - Law and Dartois prescriptions

caitlynhardiman:
Type of PR: 
Added in separate modules for running and testing the temperature stratified equation of state, now also includes Dartois et al. (2003) prescription which matches mid plane temperature

Description:
Added in separate modules for running and testing the EOS, and modified the Makefile to include these, the setup_disc routine to contain relevant parameters, and the eos.F90 and test_eos.f90 files to call to the new functions

Testing:
Passed entire test suite and eos test suite

Did you run the bots? no


danieljprice:
test of stratified eos is failing when compiled with DEBUG=yes

PR #309, opened 2022-07-22
Fixed phantom2hdf5 not compiling

s-neilson:
Type of PR: 
Fixed some files that were preventing phantom2hdf5 from being compiled.

Testing:
I successfully compiled phantom2hdf5 and was able to use it to turn PHANTOM dumps into usable .hdf5 files

Did you run the bots?
Yes.


danieljprice:
thanks for the bug fixes!

PR #314, opened 2022-08-15
timestep+ptmass+stability

jameswurster:
Type of PR: 
Bug fix & modification to existing code

Description:
timestep: debugged so that dtmax now properly decreases as density rapidly increases; can now capture the second collapse of sphereinbox as intended.
ptmass: added columns to SinkCreate, which is a useful output file for debugging, but still off by default; will now correctly print the number of neighbours and not neighbour candidates
stability: will now exit Phantom with a warning if hdivbb ave or max is greater than a given threshhold. Although not a true conservation check, being included in that section for simplicity/consistency.  This check can be overridden by setting hdivbbmax_max > 1.5, although the warning suggests 512.
Bots:  ran these and there were apparently a lot of changes, suggesting they are not frequently run... 

Testing:
Ran multiple toy & non-toy tests to verify accuracy of changes

Did you run the bots? yes

### DIFF
--- a/src/main/checkconserved.f90
+++ b/src/main/checkconserved.f90
@@ -23,6 +23,7 @@ module checkconserved
  real, public :: etot_in,angtot_in,totmom_in,mdust_in(maxdusttypes)
 
  public :: init_conservation_checks, check_conservation_error
+ public :: check_magnetic_stability
 
  private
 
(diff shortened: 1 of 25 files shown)


src/main/checkconserved.f90
```diff
@@ -123,5 +124,32 @@ subroutine check_conservation_error(val,ref,tol,label,decrease)
...
 
+ if (hdivbbmax_max < 1.1) then
+    ! In this regime, we assume the user has not modified this value,
+    ! either by choice or by being unaware of this.  This warning will
+    ! appear in this case.
+    if (hdivBB_xa(1) > 100 .or. hdivBB_xa(2) > 0.1) then
+       ! Tricco, Price & Bate (2016) suggest the average should remain lower than 0.01,
+       ! but we will increase it here due to the nature of the exiting the code
+       ! The suggestion of 512 was empirically determined in Dobbs & Wurster (2021)
+       call fatal('evolve','h|divb|/b is too large; recommend hdivbbmax_max = 512; set >1.2 to suppress this message.')
```
danieljprice, line 148: maybe this should be changed to respond to I_WILL_NOT_PUBLISH_CRAP = yes, similar to the conservation checks?

jameswurster: I thought about that as I was implementing this.  Since we can change a parameter for added stability, I thought the current method would encourage users to rerun it with shorter timesteps/higher stability in hopes that it would complete without exiting.  
When we exit via one of the other methods, the error is fatal and aside from possibly changing unit, there is little that can be done to prevent the exit aside from the environment variable.  
Please feel free to change it if you disagree.

PR #325, opened 2022-09-13
reorganization of cooling modules

lsiess:
Type of PR: 
Bug fix + new physics + modification to existing code 

Description:
- complete reorganization of cooling modules
- implementation of ray tracing routines (to be tested)
- implementation of triple system setup (in setup_wind)
- development of the analysis_cooling module to test cooling schemes
- new cooling prescriptions in cooling_functions (devel)
- bug fixes (cooling & dust_formation mainly)

Testing:
- nucleation is being heavily tested in CE simulations
- cooling & raytracing need to be further tested

Did you run the bots? no


danieljprice:
thanks for the contributions, I have merged the PR so codes are up to date. Obviously some of the stuff in analysis_cooling really belongs in test_cooling.f90 so it gets run as part of the integration testing. This can be done later  but it does mean we are  not auto-testing the cooling modules at the moment.

lsiess:
On Tue, 20 Sep 2022 07:52:20 +0000
SIESS Lionel ***@***.***> wrote:

> On Tue, 20 Sep 2022 00:41:08 -0700
> Daniel Price ***@***.***> wrote:
> 
> > thanks for the contributions,   
> thanks to you too !
> 
> > I have merged the PR so codes are up to date. Obviously
> > some of the stuff in analysis_cooling really belongs in test_cooling.f90 so it gets
> > run as part of the integration testing. This can be done later  but it does mean we
> > are not auto-testing the cooling modules at the moment.  
> absolutely. Cooling is under development and things will be moved and reorganized later.
> 
> I have another question with the shock setup. In the .setup on the left and right we
> specify the densities and the pressures (pr ?) and in the .in file the temperature T0
> and gamma. All these values cannot be set independently as they are related via the
> (perfect gas) EOS. Am i missing something ?

ok, I got this one. T0_value is consistent with the values of P and rho specified in .setup. It is output in the .in file during the phantomsetup step.
However, if afterwards, one changes T0_value in the .in file, then I don't know what this means but it is not going to be good...
I would suggest to store this value in the dump header so it cannot be modified. What do you think ?


Any thoughts about the value of alphau in my previous email ?

cheers
Lionel

lsiess:
On Tue, 20 Sep 2022 00:41:08 -0700
Daniel Price ***@***.***> wrote:

> thanks for the contributions, 
thanks to you too !

> I have merged the PR so codes are up to date. Obviously
> some of the stuff in analysis_cooling really belongs in test_cooling.f90 so it gets run
> as part of the integration testing. This can be done later  but it does mean we are
> not auto-testing the cooling modules at the moment.
absolutely. Cooling is under development and things will be moved and reorganized later.

I have another question with the shock setup. In the .setup on the left and right we specify the densities and the pressures (pr ?) and in the .in file the temperature T0 and gamma. All these values cannot be set independently as they are related via the (perfect gas) EOS. Am i missing something ?

Thanks

PR #327, opened 2022-09-20
(phantomsetup) Bug fix when using MPI

jameswurster:
Type of PR: 
Bug fix

Description:
As per tbdrstl on Slack, this is a bug patch when running phantomsetup using MPI.  @danieljprice , is there a reason that init_mpi and associated commands are so far down in phantomsetup?  The other patch would be to move the init_mpi commands to the top above set_io_unit_numbers, but I am unsure if this would cause other issues... 

Testing:
Compiled & test ran phantomsetup

Did you run the bots? no

### DIFF
--- a/src/setup/phantomsetup.F90
+++ b/src/setup/phantomsetup.F90
@@ -57,7 +57,7 @@ program phantomsetup
  real                        :: time,pmassi
  logical                     :: iexist
 
- nprocs = 1    ! for MPI, this is not initialised until init_mpi, but an initialised value is required for init_part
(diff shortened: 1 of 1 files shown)


jameswurster:
same error after re-launching tests....
Doesn't look like its a result of my one-line commit...

jameswurster:
Not sure what happened...
the final commit reflects the initial commit in terms of actual coding changes...

How often does this test suite mess up?  It seems to happen to my PRs on an annoyingly regular basis...

PR #330, opened 2022-09-28
Made compiling with gfortran on aarch64 processors possible.

s-neilson:
Type of PR: 
Bug fix

Description:
PHANTOM would previously not compile on computers that used aarch64 CPUs (at least when SYSTEM=gfortran) so I fixed the issue that was causing this (which was that mcmodel=medium is not a valid option for gfortran on aarch64 processors).

Testing:
I successfully compiled and ran the test suite on two computers that used aarch64 processors where SYSTEM=gfortran.

Did you run the bots? yes


danieljprice:
I *think* it is safe to remove the -mcmodel=medium flag in general. This used to be necessary to make sure the code was compiled in 64 bit (thus not limiting the absolute binary size to 2Gb) but I think 64 bit is default for most operating systems these days anyway?

PR #346, opened 2023-01-30
(mcfost) Fixing makefile

cpinte:
Type of PR: 
Bug fix 

Description:
Missing mcfost directory in include

Testing:
phantom now compiles with mcfost library

Did you run the bots? no


danieljprice:
changes were made here so that things work if libmcfost.a and mcfost2phantom.mod are installed to standard locations (/usr/local/lib) as would be done if libmcfost installed via a package manager (e.g. homebrew). Idea was to try to get an automated test of phantom-mcfost linking happening which requires packaging. Hence we should check this still works before merging... and ideally implement a build test in the actions

danieljprice:
just to update, this needs to be fixed at the mcfost end, ideally mcfost should copy mcfost2phantom.mod into the include/ directory after compiling, similar to how it would be copied to a system include directory if mcfost was installed to a system location

PR #350, opened 2023-02-07
Adding two addition input options with MCFOST=yes

sahl95:
Type of PR: 
other

Description:
Added two additional input options when compiling with MCFOST=yes
- mcfost_keep_part -- Fraction of particles to keep for MCFOST
- ISM -- ISM heating : 0 -> no ISM radiation field, 1 -> ProDiMo, 2 -> Bate & Keto

Testing:
I compiled with MCFOST=yes and ran a quick test simulation

Did you run the bots? no


danieljprice:
thanks Sahl!

PR #383, opened 2023-03-16
(radiation) make cv_type runtime option and set rad=0 in relax_star when using radiation

themikelau:
Type of PR: 
Bug fix

Description:
- Using relax_star with do_radiation = .true. requires setting rad=0.
- Use integer runtime option cv_type to specify method to calculate mean molecular weight in get_1overmu and in get_cv. For cv_type = 0, use mu = gmw and cv = Rg/((gamma-1.)*gmw*unit_ergg). For cv_type = 1, use MESA tables. 

Testing:
Set up radiating star.

Did you run the bots? yes/no
no

### DIFF
--- a/src/main/eos.F90
+++ b/src/main/eos.F90
@@ -1026,17 +1026,17 @@ end function get_mean_molecular_weight
 !  return cv from rho, u in code units
 !+
 !---------------------------------------------------------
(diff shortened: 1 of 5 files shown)


danieljprice:
thanks Mike

Issue #391, opened 2023-03-28
mass loss rate in wind is not equal to the value specified in the input file


danieljprice:
Using SETUP=wind, there appears to be a factor of two offset between the actual mass loss rate (mass injected in a simulation over a period of time; e.g. by measuring decrease in sink particle mass) and the desired wind_mass_rate in the input file. 

From experiments with Taissa Danilovich the two numbers appear to be discrepant by exactly a factor of 2, as in actual Mdot is a factor of two higher than the desired Mdot.

Would be good to get to the bottom of this.

danieljprice:
What was the cause?

lsiess:
On Thu, 06 Apr 2023 03:52:07 -0700
Daniel Price ***@***.***> wrote:

> What was the cause?
> 

mass was removed twice. In inject_particles we had the following sequence:

 do i=inner_boundary_sphere,outer_sphere,-1

.....

       ! update the sink particle mass
       if (nptmass > 0 .and. wind_emitting_sink <= nptmass) then
          xyzmh_ptmass(4,wind_emitting_sink) = xyzmh_ptmass(4,wind_emitting_sink) - mass_of_spheres
       endif
       print '(" ##### eject sphere ",4(i4),i7,9(1x,es12.5))',i,inner_sphere,nboundaries,&
            outer_sphere,npart,time,local_time,r/xyzmh_ptmass(iReff,1),v,u,rho
    endif
    !cs2max = max(cs2max,gamma*(gamma-1)*u)
 enddo
 ! update sink particle properties
 if (nptmass > 0 .and. wind_emitting_sink <= nptmass) then
    mass_lost = mass_of_spheres * (inner_sphere-outer_sphere+1)
    xyzmh_ptmass(4,wind_emitting_sink) = xyzmh_ptmass(4,wind_emitting_sink) - mass_lost
    if (pulsating_wind) then
       inner_radius = wind_injection_radius + deltaR_osc*sin(omega_osc*time)
       !v2 xyzmh_ptmass(5,wind_emitting_sink) = (inner_radius**3-dr3)**(1./3.) !accretion radius
       xyzmh_ptmass(5,wind_emitting_sink) = inner_radius
    endif
 endif

As you see, mass was removed twice from the wind emitting sink particle. I suspect this error was introduced in a merge. 
It is not the first time I am discovering unpleasant mistakes like this after a merge... 

Cheers
Lionel

danieljprice:
[...]

PR #392, opened 2023-03-28
Wind bug fixes

lsiess:
Type of PR: 
Bug fixes

Description:
Fix bugs, clean code and improve comments

- fix issue #[391 mass loss rate in wind is not equal to the value specified in the input file](https://github.com/danieljprice/phantom/issues/391)
- fix critical error in the calculation of the dust temperature (ray-tracing routines)
- in wind_setup, if wind_temperature = 0, sets wind_temperature = Tstar
- improve comments in ray-tracing routines

Testing:
ran simulations with all ray-tracing options in both single and binary star systems

Did you run the bots? yes


src/main/inject_wind.F90
```diff
@@ -469,10 +469,6 @@ subroutine inject_particles(time,dtlast,xyzh,vxyzu,xyzmh_ptmass,vxyz_ptmass,&
        endif
        !initialize dust temperature to star's effective temperature
        if (isink_radiation > 0) dust_temp(npart+1:npart+particles_per_sphere) = xyzmh_ptmass(iTeff,wind_emitting_sink)
-       ! update the sink particle mass
-       if (nptmass > 0 .and. wind_emitting_sink <= nptmass) then
-          xyzmh_ptmass(4,wind_emitting_sink) = xyzmh_ptmass(4,wind_emitting_sink) - mass_of_spheres
-       endif
```
danieljprice, line 475: this is the only line I am concerned about, it means that we no longer conserve mass when emitting stellar winds, that is the mass loss is not taken from the sink particle mass. 

Can you explain why this has been removed?

src/main/deriv.F90
```diff
@@ -195,22 +197,51 @@ subroutine derivs(icall,npart,nactive,xyzh,vxyzu,fxyzu,fext,divcurlv,divcurlB,&
     call get_growth_rate(npart,xyzh,vxyzu,dustgasprop,VrelVf,dustprop,ddustprop(1,:))!--we only get ds/dt (i.e 1st dimension of ddustprop)
  endif
```
danieljprice, line 199: just a general comment here, I think it would be good to move the various options here into a daughter routine called get_dust_temperature () where all the various options are handled inside, rather than having them clutter the derivs routine.

I'm ok to merge this but would be nice to clean it up sometime soon..

danieljprice (commented):
looks good. Please can you explain why the mass loss accounting from the sink particle was removed?

Also what was the reason for #391 as to why the Mdot was off by a factor of two?

danieljprice:
Hi Lionel & Mats,

 I fixed the build failure. I cannot update this pull request directly, so instead I submitted the changes to your fork via a pull request, which when accepted should automatically update (and fix) this one. The only change is to get rid of the compiler warnings, plus I merged the latest code and fixed the conflict in the AUTHORS file.
 
Cheers,

Daniel

lsiess:
Hi Daniel

I just merge lsiess-master with my master branch and push it. 

Thanks for your help with the compiler option!

Cheers
Lionel

PR #393, opened 2023-03-28
Bug fixes related to star and binary star setup

themikelau:
Type of PR: 
- Bug fix

Description:
- `set_star_thermalenergy` now takes `npts` as an input so that density and pressure arrays can be interpolated in their valid range. This fixes the bug where array interpolation gives den = pres = 0, e.g., when setting up a polytrope with `relax_star = .false.`.
- Reorganisation of `moddump_binary` that fixed some bugs. Now tested that setup = 1, 8 options work for producing 6 possible types of binary stars: 1) star+star, 2) sink-star+star, 3) star+sink-star, 4) sink-star+sink-star, 5) sink-star+sink, 6) star+sink.


danieljprice:
build failure is just a line truncation error blocking compilation of the moddump

also, could we think about writing a unit test for the moddump_binary to make sure these bugs do not reoccur?

PR #395, opened 2023-03-30
Hierarchical systems

si-mon-jinn:
Type of PR: 
New setup option

Description:
Add the possibility to build a generic hierarchical sink particle system.

Testing:
Checked by hand that everything is working.

Did you run the bots? yes


danieljprice:
the final build failure here is just in
```
make setup SETUP=wind
```
giving the error:
image

PR #406, opened 2023-04-23
(radiation) control flux limiter with switch; read additional radiation quantities from dumps

themikelau:
Type of PR: 
- Bug fix + feature

Description:
- Explicit radiative diffusion now selects whether or not to use the (Levermore & Pomraning) flux limiter depending on the input option `limit_radiation_flux`. Previous behaviour was to always use an optically-thick diffusion coefficient (lambda->1/3), which is a bug. Can @danieljprice check if this change is safe for implicit radiative diffusion?
- readwrite_dumps modified to also read grad(E) and lambda from a dump that used radiation.


danieljprice:
all looks good to me

PR #407, opened 2023-04-24
(set_star) shifted all the work in setup_star into set_star routine

danieljprice:
Type of PR: 
cleanup / structural changes to set_star

Description:

The aim of this change is to be able to perform all the functionality of setup_star from any setup routine in phantom. This is with the intention of making several "two-step" setup-and-moddump procedures into a single, streamlined procedure. For example, the TDE setup could add a star, relax the star and place it into orbit without needing the moddump procedure. Similarly, this change allows implementation a binary, triple or hierarchical star setup with multiple stars read from MESA tables.

The changes to the code at the moment are purely structural, there is no change in functionality:

- added new set_star routine/module that does all the steps in setup_star
- set_star provides routines to read/write/ask for all the options needed to setup and relax a single stellar model
- teased apart the options that would be per-star from the equation of state options that are per-setup

Testing:
Testing on AGB stellar model with sink particle core, ensuring setup procedure proceeds as previously

Did you run the bots? no

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -659,7 +659,7 @@ SRCSETUP= prompting.f90 set_dust_options.f90 \
           density_profiles.f90 set_slab.f90 set_disc.F90 \
           set_cubic_core.f90 set_fixedentropycore.f90 set_softened_core.f90 \
           set_star_kepler.f90 set_star_mesa.f90 \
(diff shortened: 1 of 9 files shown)


themikelau (approved):
Looks okay.

Issue #341, opened 2022-12-05
Phantomanalysis with phantom+mcfost


emborchert:
When running phantom+mcfost (SETUP=mcfost), it is currently not possible to run phantomanalysis with the default simulation setup. There seems to be something doubling up with the mcfost files.

Running phantomanalysis, the following output is created:

```
$make analysis ANALYSIS=analysis_binarydisc.f90
cd /home/eborcher/phantom/; make SETUP=mcfost RUNDIR=/home/eborcher/runs/solar/paper/test01 KROME= phantomanalysis; cd -; cp /home/eborcher/phantom//bin/phantomanalysis .
make[1]: Entering directory `/home/eborcher/phantom'
make[2]: Entering directory `/home/eborcher/phantom/build'

Compiling Phantom v2022.0.1 for ifort system...........

make: WARNING: value of SETUP = mcfost is obsolete...
=> setting SETUP = disc

Using options for disc setup

Using cubic kernel
Flags set for DOUBLE PRECISION
Compiling in PARALLEL (OpenMP)
Preprocessor flags are -DMCFOST -DDISC_VISCOSITY -DIND_TIMESTEPS -DLIVE_ANALYSIS
Fortran flags are -O3 -inline-factor=500 -mcmodel=medium -shared-intel -warn uninitialized -warn unused -warn truncated_source -no-wrap-margin -fPIC -qopenmp -r8
[...]

PR #412, opened 2023-05-05
GitHub actions workflow to check phantom+mcfost

### LINKED ISSUE #379: need GitHub actions check for phantom+mcfost
as part of the continuous integration workflow, we should test that linking phantom with mcfost works correctly and is not broken

danieljprice:
Type of PR: 
continuous integration testing, closes #379 

Description:
this is a GitHub action to compile phantom and link it against MCFOST, as needed for live phantom+mcfost simulations
the idea is to check if the API gets broken for whatever reason (could either be at the phantom or on the mcfost end).

Testing:
Will be tested in the actions workflow

Did you run the bots? no

Did you update relevant documentation in the docs directory? yes, the action follows recommended installation procedure for homebrew installation of mcfost in the mcfost help page

### DIFF
--- a/.github/workflows/mcfost.yml
+++ b/.github/workflows/mcfost.yml
@@ -23,21 +23,22 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+    - name: install gfortran
+      run: brew install gfortran
+
+    - name: soft link gfortran and check version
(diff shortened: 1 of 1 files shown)


danieljprice:
the failure here is due to #341 and needs to be fixed by pull request to MCFOST

PR #418, opened 2023-05-12
Hierarchical systems

si-mon-jinn:
Type of PR: 
Bug fix 

Description:
Fixed read function usage by CHESS

Testing:
Tested hierarchical system setup with discs both with gfortran and ifort

Did you run the bots? no


danieljprice:
the failure here is an variable-used-without-being-initialised error flagged when running phantomsetup using ifort with DEBUG=yes:

image

Issue #46, opened 2020-08-07
issues with phantom2hdf5


danieljprice:
the changes on this are an improvement, so thanks for the efforts. I will merge to avoid the broken phantom2hdf5 but there are still a number of issues here:
1. there is currently lots of manual maintenance required to keep the formats in sync
2. the header quantities have to be manually written and read in hdf5 and the struct would have to change every time another quantity is added to the header
3. array quantity read/write is also duplicated, this needs to be defined in only one place in the source code
4. there are now at least 5 fortran source code files required to be compiled just to read and write a file

As mentioned, I think there is a much better way to fix (3), which is the ultimate problem here, as per #43

_Originally posted by @danieljprice in https://github.com/danieljprice/phantom/pull/39#issuecomment-670277737_

danieljprice:
the bigger issue with this utility, as per #70 is that format conversion should really not be done via the phantom read/write dump utilities, since:

1) the read_dump reads only the necessary information to continue a calculation (and it would be unwise to change this)
2) what is written by write_dump depends on how phantom was compiled

both of these issues are problematic if the desire is to simply convert a format, since it results in information loss.

So, phantom2hdf5 either needs a rewrite or should be made obsolete in favour of a better solution (e.g. splash or pysplash performing the conversion)

PR #437, opened 2023-06-16
AOCC compiler fix

Fitz-Hu:
Type of PR: 
Bug fix 

Description:
Add flag when using aocc so it fixes the bug automatically

Testing:
ran make test

Did you run the bots? no


danieljprice:
fix looks good, I will merge for now but eventually would be good to eliminate the need for this. It is related to #46, we should not need procedure pointers to switch the file format as we can deprecate the direct-write hdf5 option now that Sarracen exists and replace it with an hdf5 conversion utility

PR #438, opened 2023-06-21
Few and minor cleanup to modules related to radiation 

themikelau:


themikelau:
I'm just failing on "Failed to connect to www.minorplanetcenter.net". I assume that's not a problem.

PR #441, opened 2023-06-29
Omp radiation

themikelau:
Merge parallel sections in implicit radiation code.


danieljprice:
dti can just be private I think as it would now be defined on all threads

PR #443, opened 2023-07-03
Hyperbolic and parabolic setups in set_binary

### LINKED ISSUE #302: Impossible to create a stellar collision or a fly by between 2 stars.
I was trying to set up a binary system. However, I have agreed that the options for setting the orbit are very limited. For example, it is impossible to impose an eccentricity greater than 1, which makes it impossible to simulate a stellar collision between two desired stars (without necessarily merging).

Would it be nice if:
-Eccentricity greater than 1 could be entered (and more changes in the orbit of the two stars in general)
-The possibility of rotating the star during its relaxation or evolution could be added. The result, during fusion or collision, would be very different if they had their own rotation.
Thank you for your work.

danieljprice:
Type of PR: 
new physics; resolves #302

Description:
Added the ability to handle parabolic and hyperbolic orbits in the set_binary routine, which resolves #302. This does not replace the functionality in set_flyby because by default this routine uses the true anomaly to set the position in the orbit. which is not terribly helpful for unbound orbits. However, it allows one to simply set e >= 1 in any setup calling set_binary and things will at least work.

Testing:
manual setup of point mass binary using SETUP=binary:
```
~/phantom/scripts/writemake.sh binary > Makefile
make setup
make
-> edit .setup file to give e=1 or e>1 and play with angles
```

Did you run the bots? no


Francyrad:
I am infinitely grateful! I will run a test immediately and compare the results with StarSmasher

PR #449, opened 2023-07-06
Remove leftover ifdef statements to fix stirring bug

ttricco:
Type of PR: 
Bug fix 

Description:
Turbulent stirring had the ifdef STIR_FROM_FILE removed in commit 7c86f4e. One set of ifdef statements were missed. This would cause the turbulent stirring to try to not stir from file. Now all STIR_FROM_FILE ifdef statements are removed, instead using the logical parameter.

Testing:
Before and after test on SETUP=turb. Previously would error the first time the code tried to force.

Did you run the bots? N/A

Did you update relevant documentation in the docs directory? N/A


danieljprice:
thanks Terry, we really should have a test that picks up this kind of thing...

Issue #450, opened 2023-07-06
add automated test that checks that setups run correctly


danieljprice:
we already have a test that checks that phantomsetup runs, accepts "enter" to any questions asked, successfully completes and writes a .in file. 

However, there are a few issues that still slip through:

```
              thanks Terry, we really should have a test that picks up this kind of thing...
```
_Originally posted by @danieljprice in https://github.com/danieljprice/phantom/issues/449#issuecomment-1624401595_

Suggestion would be to run not only phantomsetup in the actions tests, but perhaps also run phantom on the .in file, maybe just with nmax=0 to ensure that it goes through one loop of derivs successfully?

danieljprice:
confirmed that if I run:
```
~/phantom/scripts/writemake.sh turb
make setup
make
./phantomsetup turb
[press enter]
./phantom turb.in
```
then with the broken code (pre #449) I get:
```
 Reading forcing.dat in /Users/dprice/phantom/data/forcing/ (from PHANTOM_DIR setting)
 ERROR: cannot write to /Users/dprice/phantom/data/forcing/, writing to current directory
--------------------------------------------------------------------------------
  Downloading forcing.dat from https://users.monash.edu.au/~dprice/phantom//data/forcing/
--------------------------------------------------------------------------------
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13.7M  100 13.7M    0     0  2697k      0  0:00:05  0:00:05 --:--:-- 2994k
--------------------------------------------------------------------------------
 DOWNLOADED forcing.dat TO forcing.dat
read_stirring_data_from_file: read new forcing pattern, stepinfile =      0 , time =         0.000000E+00 , time in stirring table =         0.000000E+00
  Initialized          112  modes for stirring.
  spectral form        =            1  (Parabola)
  solenoidal weight    =    1.0000000000000000     
  st_solweightnorm     =    1.2247448713915889     
  stirring energy      =    8.0000000000000000     
  autocorrelation time =    5.0000000000000003E-002
  minimum wavenumber   =    6.2800000000000002     
  maximum wavenumber   =    18.899999999999999     
  original random seed =            1
  bulk motion correction is OFF
Fortran runtime error: Array rank of GET is not 1.

Error termination. Backtrace:
#0  0x1031c5aae
#1  0x1031c67a5
#2  0x1031c6be8
#3  0x1034218d1
#4  0x100de25f5
[...]

Issue #55, opened 2020-08-18
remove #ifdefs from phantom


danieljprice:
#ifdefs look horrible in the code, and make testing all possible compile-time combinations very difficult. We should avoid introducing any new #ifdefs, preferencing use of the logical flags from the config module where possible, e.g. if (mhd), if (radiation) etc.

Many of these could also be made into runtime options, since they are mainly there to determine whether memory is allocated for certain arrays. This would need a bit more thought.

For now, we should try to keep as many files as .f90 as possible, rather than .F90, to disallow the use of ugly and unnecessary #ifdef statements. The only allowed use case for #ifdefs should be for linking with external libraries, e.g. MPI, or when the performance *really* matters (e.g. use of #ifdef PERIODIC in dens, force or kdtree)

danieljprice:
Current status:
```
Number of source files (.f90, .F90):      442
Number of .f90 files               :      380
Number of .F90 files               :       62
Number of unique #ifdef statements :       48
Number of total  #ifdef statements :      526
```

danieljprice:
```
Number of source files (.f90, .F90):      473
Number of .f90 files               :      413
Number of .F90 files               :       60
Number of unique #ifdef statements :       48
Number of total  #ifdef statements :      500
```

danieljprice:
```Number of source files (.f90, .F90):      476
Number of .f90 files               :      427
Number of .F90 files               :       49
Number of unique #ifdef statements :       41
Number of total  #ifdef statements :      413
```

danieljprice:
```
Number of source files (.f90, .F90):      475
Number of .f90 files               :      430
Number of .F90 files               :       45
Number of unique #ifdef statements :       33
Number of total  #ifdef statements :      393
```

PR #446, opened 2023-07-05
Add wind tests to the test suite (non dusty winds)

lsiess:
Type of PR: 
new test_wind.f90 routine

Description:
Provide a test suite for non dusty  winds from a single star

Testing:
I ran make test and passed all the tests

Did you run the bots? yes

Did you update relevant documentation in the docs directory? no

PR #451, opened 2023-07-06
Automated test that ./phantom blah.in succeeds after running phantomsetup

danieljprice:
Type of PR: 
Bug fix / new test

Description:
See #450 and #449: In the buildbot we now check not only that ./phantomsetup blah succeeds, but also that ./phantom blah.in successfully runs for at least one timestep

Testing:
Checked that this works with SETUP=turb, will be checked fully in the GitHub actions

UPDATE: a LOT of the setups turned out to be broken when run this way, particularly with DEBUG=yes. Slowly working through the backlog of complaints from the actions... tidying up a few other things (especially #55) as I go along

Did you run the bots? yes


danieljprice:
@lsiess three of the remaining test failures are from "wind", "ismwind" and "radwind", which I think will be fixed when #446 is merged by the looks of your code (i.e. your new set_defaults routine)

...so please can you get #446 to pass so it can be merged...?

lsiess:
The problem with the test setups is that you have to compile and link with `utils_binary.f90 set_binary.f90 inject_wind.f90` because they are needed for `test_wind`. I found an ugly way of doing it in the `Makefile` with this
```
ifndef SRCINJECT
    SRCINJECT=utils_binary.f90 set_binary.f90 inject_wind.f90
endif
```
but I am sure you will find a more elegant way of doing it. 
cheers

danieljprice:
got it, I think I fixed this bit at least

PR #474, opened 2023-10-12
Signal speed used in tcour should include radiation pressure

themikelau:
Add contribution of radiation pressure to signal speed when using radiation (e.g., Eq 10 in https://arxiv.org/abs/1102.1216). We should have cs = (5/3*Pgas/rho + 4/3*Prad/rho)^1/2. Otherwise, tcour may be underestimated.


danieljprice:
this is a good idea, however we should test before merging as it would also increase the shock viscosity. Best problem for this would be the radiation shock tube (radshock) although we'd need to do it in a regime where radiation pressure is significant to see what difference this makes to the shock capturing (and indeed the stability)

PR #481, opened 2023-11-02
Set units for shock setup

themikelau:
- Interactive setup now prompts for mass and length units
- Units now written into and read from setup file

### DIFF
--- a/src/setup/setup_shock.F90
+++ b/src/setup/setup_shock.F90
@@ -438,7 +438,7 @@ subroutine choose_shock (gamma,polyk,dtg,iexist)
  zright = 0.0
  const  = sqrt(4.*pi)
 
- call set_units_interactive(gr)
+ if (do_radiation) call set_units_interactive(gr)
 
 !
 !--list of shocks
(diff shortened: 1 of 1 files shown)


danieljprice (commented):
@themikelau can you explain the motivation for this change? In general we should NOT set units for shock tube problems, since they are scale-free problems. The only way units change anything is if you have radiation or other physics that depends on the actual temperature. Hence my strong preference would be to only set units for the cases where this is useful i.e. if (do_radiation). 

Otherwise simple test problems become confusing to the user, especially since splash will present things in physical units by default if they are set...

themikelau:
@danieljprice  Thanks for the review. Indeed I added this for running a radiation-dominated shock. I will add `if (do_radiation)` around my changes.

danieljprice:
I have raised an issue on the slow actions, #484 ,  looking into it but this is currently blocking merge (I have rerun all tests and same issue). Affecting both currently open pull requests, so it's nothing from the proposed merges, either something from a previous merge or something changed on the actions runners...

PR #483, opened 2023-11-16
Add v_esc option for .divv files

miguelglezb:
Type of PR: 
Bug fix / new physics / modification to existing code / other

Description:
Give a short description of the change.

Testing:
Describe how you have tested the change

Did you run the bots? yes/no


danieljprice:
The test slowdowns here are real but definitely not caused by this p-r. Seems like the sink particle tests with MPI and openMP have got really slow for some reason... I can reproduce this on my laptop, looking into it...

PR #513, opened 2024-02-22
Libphantom-amuse

rieder:
Type of PR: 
new code: library for using Phantom with AMUSE

Description:
- introduce a `write_files` option, defaults to `.true.`
- make write/open/close statements conditional on `write_files`
- introduce `utils/libphantom-amuse.F90`, with helper subroutines for the AMUSE interface

Testing:
WIP

Did you run the bots? yes/no


src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+ character(len=120) :: infile,logfile,evfile,dumpfile
+
+ id = 0
+
+ call init_mpi(id,nprocs)
+ call allocate_memory(int(maxp_hard,kind=8))
+ call set_io_unit_numbers()
+ call initialise()
+ call amuse_set_defaults() ! replaces reading infile
+ call set_units(dist=1 * au,mass=1.*solarm,G=1.)
```
danieljprice, line 29: one option here is to send strings for the units, then call select_unit(mass_unit,umass,ierr) to get the unit from the string

danieljprice: this is inside setup/set_units.f90

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+ 
+end subroutine amuse_initialize_code_new
+
+subroutine amuse_initialize_code()
```
danieljprice, line 33: can we delete this if obsolete?

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+    ibinnow       = 0
+#else
+    dtcourant = huge(dtcourant)
+    dtforce   = huge(dtforce)
+#endif
+
+    !call amuse_initialize_wind()
+end subroutine amuse_initialize_code
+
+subroutine amuse_set_phantom_option(name, valstring, imatch)
```
danieljprice, line 85: this could in principle be done in a more automated way, if we were to read the infile like setup files... this is on my list (should create an issue on this to get this task done)

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+    print*, "COMMIT_PARTICLES, calling amuse_init_evol"
+    !call amuse_init_evol()
+    call amuse_evol(.true.)
+end subroutine amuse_commit_particles
+
+subroutine amuse_recommit_particles()
+    use deriv, only:derivs
+    use part, only:npart,xyzh,vxyzu,fxyzu,fext,divcurlv,divcurlB,Bevol,dBevol,&
+            dustprop,ddustprop,dustfrac,ddustevol,dens,drad,radprop,dustevol,&
+            eos_vars,metrics,pxyzu,rad
```
danieljprice, line 337: I think you can just call get_derivs_global() here which calls derivs, then the interface to derivs would not need updating. This can be found in the derivs module and is used in the test suite

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+    implicit none
+    integer, intent(out) :: npart_out
+    logical, intent(in)  :: nodisabled
+
+    integer :: i
+
+    if (nodisabled) then
+        npart_out = 0
+        do i=1,npart
+            if (xyzh(4,i)  >  0.) then
```
danieljprice, line 371: this condition should be a call to isdead_or_accreted(xyzh(4,i)) for consistency with the other code

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+   ! A few changes to Phantom's defaults
+   call amuse_set_gamma(1.)
+   call amuse_set_ieos(1)
+   write_files=.false.
+
+end subroutine amuse_set_defaults
+
+! This initialises things. This really should only be called once, before the first step.
+subroutine amuse_evol(amuse_initialise)
+ use io,               only:iprint,iwritein,id,master,iverbose,&
```
danieljprice, line 395: this should be removed and replaced by a call to evol but with some flags needed for amuse to behave correctly

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+
+subroutine amuse_delete_particle(i)
+    use part, only:kill_particle,xyzmh_ptmass
+    integer, intent(in) :: i
+    integer :: j
+    if (i == abs(i)) then
+        call kill_particle(i)
+    else
+        j = -i
+        ! Sink particles can't be killed - so we just set its mass to zero
```
danieljprice, line 1150: sink particles can be killed by giving them a negative mass, suggest xyzmh_ptmass(4,i) = -abs(xyzmh_ptmass(4,i))

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+        xyzmh_ptmass(4,j) = 0
+    endif
+end subroutine
+
+subroutine amuse_get_unit_length(unit_length_out)
```
danieljprice, line 1155: should not be a need for these routines, as long as set_units() is called during the amuse_initialize_code_new routine, and if you can send strings to set different options

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+    integer, intent(in) :: i
+    double precision, intent(in) :: u
+    if (maxvxyzu >= 4) then
+        vxyzu(4, i) = u
+    endif
+    ! Changing temperature -> better use a small cooling step
+    dtextforce = 1.e-8
+end subroutine
+
+subroutine amuse_evolve_model(tmax_in)
```
danieljprice, line 1748: see previous, should be just a call to evolve

src/utils/libphantom-amuse.F90
```diff
@@ -0,0 +1,2306 @@
...
+end subroutine
+
+subroutine amuse_get_udist(udist_out)
+    use units, only:udist
+    implicit none
+    double precision, intent(out) :: udist_out
+    udist_out = udist
+end subroutine
+
+! End of Getters
```
danieljprice, line 2275: please delete commented out/ obsolete code

danieljprice (changes requested):

src/utils/libphantom-amuse.F90
rieder: done

src/utils/libphantom-amuse.F90
rieder: changed

src/utils/libphantom-amuse.F90
rieder: done

src/utils/libphantom-amuse.F90
rieder: done

src/utils/libphantom-amuse.F90
rieder: Done

src/utils/libphantom-amuse.F90
rieder: done

rieder:
Ready to merge for me.

rieder:
Made some changes to `libphantom-amuse` for consistency with AMUSE, which uses 32-bit integers for particle indices instead of the 64-bit integers Phantom uses (see https://github.com/amusecode/amuse/issues/1077). The interface will now stop with an Error if this would cause a problem. Also gets rid of some warnings when the AMUSE interface is compiled.

PR #522, opened 2024-04-02
New moddump that converts non-radiative dump to radiative dump

themikelau:
Tested that using moddump_LTE_to_rad and moddump_rad_to_LTE in succession recovers the same dump file to within machine precision.


danieljprice:
thanks Mike!

PR #523, opened 2024-04-03
Add the option to integrate ptmass with 4th order scheme (FSI)

Yrisch:
Type of PR: 
New integration method

Description:
Add new 4th order integration schemes for point mass evolution. At this moment, The FSI and PEFRL are implemented. 
Only the gravitational interactions are taken into account. The PEFRL is certainly not well suited for Phantom ( due to backward step). To extend the FSI to all other interactions, it is needed to compute the gradient term for each of them in the middle kick phase.  cf : Chin (2007a)  10.1103/PhysRevE.75.036701

Testing:
To test these methods, I reused the galcen setup without gas injection. Then the system of point masses is integrated during a short period and a check on the energy and angular momentum conservation is performed. A little test with 2 point masses is also done to check if the evolution correspond to a keplerian motion. 

Did you run the bots? no


src/main/config.F90
```diff
@@ -122,6 +122,13 @@ module dim
  logical, parameter :: do_radiation = .false.
 #endif
 
+ ! Regularisation method and/or higher order integrator
```
danieljprice, line 125: see #55, this really should be a runtime parameter, not a compile-time parameter. Suggest to remove the #ifdef here...

build/Makefile
```diff
@@ -274,6 +274,10 @@ ifeq ($(RADIATION), yes)
     FPPFLAGS += -DRADIATION
 endif
 
+ifeq ($(FOURTHORDER), yes)
+    FPPFLAGS += -DFOURTHORDER
+endif
```
danieljprice, line 279: same comment as below... no new ifdefs!

src/main/ptmass.F90
```diff
@@ -110,6 +111,27 @@ module ptmass
  private
 
 contains
+
+!----------------------------------------------------------------
```
danieljprice, line 115: I will open an issue on this, we should fix the kernel-generating python script to also produce this routine #523

src/main/step_leapfrog.F90
```diff
@@ -1048,6 +1053,393 @@ subroutine step_extern_sph(dt,npart,xyzh,vxyzu)
 
 end subroutine step_extern_sph
 
+!----------------------------------------------------------------
```
danieljprice, line 1056: I think we should move all step_extern routines into a separate module (step_extern.f90 ?)

Yrisch:
something wrong appends with the quintic kernel module... There is no new grad kernel

Issue #161, opened 2021-08-20
HDF5 routines are not tested


conradtchan:
Adding a build test for `HDF5=yes` would prevent issues like #137.

danieljprice:
closing this as duplicate of #46

PR #532, opened 2024-04-18
(build) add aocc installation

themikelau:
Type of PR: 
Bug fix / new physics / modification to existing code / other

Description:
Give a short description of the change.

Testing:
Describe how you have tested the change

Did you run the bots? yes/no

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,6 +54,7 @@ jobs:
       fail-fast: false
       matrix:
         system:
+          - aocc
           - gfortran
           - ifort
         batch: ${{ fromJson(needs.matrix_prep.outputs.batch) }}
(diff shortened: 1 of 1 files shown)


danieljprice:
Thanks for this Mike!

- the failure is a Fortran 2008 command (findloc) that is used in set_heirarchical_utils. This also breaks very old versions of gfortran... the workaround would be to write our own subroutine with equivalent functionality, but it seems stupid to make our own routine to reimplement a command that is already in the language...

- there is also one known "missing feature" in aocc to do with default initialisation of the procedure pointers for readwrite_dump, which are currently worked around with an #AOCC ifdef, this may cause some build failures in utilities (moddump etc) where the procedure pointers are not explicitly set (by a call to init_readwrite_dumps). Given Sarracen now exists and there are a bunch of problems maintaining the hdf5 format (#46, #161) this stuff is ripe for an overhaul (e.g.  deprecate hdf5 functionality and just provide a phantom2hdf5 converter, see #46)

themikelau:
Thanks for the clarification, @danieljprice. Good to know the test is successfully failing. I will first start with a replacement of findloc. I think it can be replaced in-line using other intrinsic functions like minloc or maxloc, without having to write new subroutines. This is worth doing since findloc hasn't been used in other places, and it is desirable to quickly implement an aocc build test before we push more aocc compiler errors.

src/setup/set_hierarchical_utils.f90
```diff
@@ -403,13 +403,13 @@ subroutine find_hier_level_orb_elem(hl_temp, hs, m1, m2, accr1, accr2, &
  !print *,'labels passing: ', trim(adjustl(hl_temp))//'1 ', m1,trim(adjustl(hl_temp))//'2 ',m2
 
  if (any(hs%labels%sink == trim(adjustl(hl_temp))//'1')) then
-    accr1 = hs%sinks(findloc(hs%labels%sink,trim(adjustl(hl_temp))//'1', 1))%accr
+    accr1 = hs%sinks( maxloc( merge(0,1,hs%labels%sink==trim(adjustl(hl_temp))//'1' ), 1) )%accr
```
danieljprice, line 406: Would suggest to at least write a helper function here like findloc_nof08 to avoid repeated code

PR #536, opened 2024-04-24
Add subsystem integration using regularization method like TTL or logH (Slow down will come in an other PR)

Yrisch:
Type of PR: 
New integration method

Description:
Add of two main new feature. First, the sub group identification. This groups can be hard binaries, hierarchical multiples that are extremely expensive to integrate in a star cluster or star formation simulation using direct summation (Gas could be in theory added into such groups but need furthermore investigation). After being identified, subgroups can be integrated using regularization algorithm like TTL or LogH. This is also done isolated from the global system. That way it drastically improves  performance  in the N² direct summation. (See Wang + 2020,Rantala + 2023 for more details)
This should be used in star cluster and star formation sim. It could also be used in other system but with extra care... (more investigation needed)
Testing:
The testing is done on a binary cluster including a full spectrum in mass ratio, eccentricity and semi of primordial binaries.
This test is done without gas at the moment but should be done with soon..
2 more tests in the ptmass test suite will also be added. One simple circular and equal mass binary  and one 3 body problem.

Did you run the bots? no


danieljprice:
Is this ready to merge?

Yrisch:
All this stuff is turned off by default. So, it should be ready to merge without breaking anything. The only thing that could be risky is the allocation of my ntpmass² matrix if nptmass is higher than 20000(already 400 mbytes with this value). But as far as I know, there is no setup with such value ?

Yrisch:
I should update the checksetup file before merging to avoid mistakes with the options...

Yrisch:
In the end the easiest and cleanest way was to remove the runtime option and set the flag to true in the setup file. After this last commit, It should be ready to merge.

Issue #538, opened 2024-05-01
last version of the code breaks the wind setup when isink_radiation=1


lsiess:
After the addition of the 4th order scheme (April 22, hash 282e51), the wind setup does not work anymore when isink_radiation = 1

Yrisch:
Hi,
For now, you can force the code to work with leapfrog (set use_fourthorder to false). It should resolve the issue temporarily. I checked the routine called for sink radiation in the new substep force routine and it needs an update to use the extrapolation method (mandatory for 4th order). The position of gas and sink particles should be slightly shifted following Chin 2007. I will try to do that in the next few days. Which test I can try to validate the new implementation ?

lsiess:
Hi,
An easy test is to run a wind simulation with an old version of Phantom that worked (e.g. 63d8aef) and compare the results with those obtained with the 4th order scheme.
To generate the code :     writemake.sh wind > Makefile
and attached are the .in and .setup files 
[wind_setup.txt](https://github.com/danieljprice/phantom/files/15187150/wind_setup.txt)
[wind_in.txt](https://github.com/danieljprice/phantom/files/15187151/wind_in.txt)

Thanks!

Yrisch:
Thanks ! I'll try to fix that soon !

lsiess:
Hi Yann

If you plot the radial velocity vr = f(r) 
Attached is what I obtain : in red old version in black new one

![image](https://github.com/danieljprice/phantom/assets/21124753/493729f4-8b82-4128-85c3-6d3654990029)

lsiess:
No, isink_radiation should only adds the radiation pressure force : kappa L/(4 pi r^2 c)

danieljprice:
yes but there is the e^(-tau) term used in some cases

lsiess:
tau should only be needed in Mat's ray tracing, i.e. if iray_resolution /= -1

Yrisch:
the issue comes from the radiation pressure force that is not accounted during the middle force (using extrapolation) with FSI... I forgot to make a patch for this part of the code. I'm working on it.

Yrisch:
Ok I made a patch, but I had to merge get_rad_accel_radiation routine with my main gas force loop. was there a specific reason that the two force was calculated separately in the previous code ?

Yrisch:
Should be fixed with #539

danieljprice:
[...]

PR #539, opened 2024-05-02
patch rad_accel_ptmass to be used with FSI

Yrisch:
Type of PR: 
Bug fix 

Description:
Patch the rad_accel_ptmass routine to be correctly used with the new FSI scheme.

Testing:

Did you run the bots? no

### DIFF
--- a/src/main/ptmass_radiation.f90
+++ b/src/main/ptmass_radiation.f90
@@ -58,16 +58,25 @@ end subroutine init_radiation_ptmass
 !  compute radiative acceleration from ALL sink particles
 !+
 !-----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


danieljprice:
this fix is pretty horrible, can we understand why it doesn't work with the acceleration computed inside the get_force routine. I would strongly prefer it to stay there if possible...

Yrisch:
It's already the case. I also agree that this part should stay in the get_force routine. The point was on where it was in this routine. Sorry for the misleading. I badly explained myself in #538.

danieljprice:
sorry I think I misunderstood the change here. I added a fix to the mcfost build failure...

PR #542, opened 2024-05-06
Wind test including radiation force + ieos = 17 + bug fixes

### LINKED ISSUE #540: Need tests for sink_radiation functionality
given #538 we should have something in the test suite that checks that the wind with sink radiation cannot be broken by code changes. Ideally a test of the wind with radiation pressure against an analytic solution?

lsiess:
Type of PR:  
Bug fix /  modification to existing code (fixes #540)

Description:
Include a new series of tests in the test_wind module :
- check total internal and kinetic energy of the particles after the test
- test wind problem when radiation force + Bowen opacity are included

Implement ieos = 17 to keep gamma constant but allow the mean molecular weight to change according to H2 formation

Bug fixes
- case ieos = 5 was restricted to simulations with nucleation. It can now be used in any setup (same for ieos=17)
- fix bug in temperature evaluation in calc_muGamma

Testing:
run simple wind models with ieos=5,17 to check that mu and gamma where not held constant

Did you run the bots? no


danieljprice:
this all looks good, remaining failure is just that the tolerance needs to be a bit higher for the internal energy:

```
 checking wind mass loss rate...........OK     [max err = 1.330E-16, tol = 2.220E-16]
 checking total internal energy.........FAILED [got  2.218E+02 should be   2.218E+02, err = 8.584E-15, tol = 5.000E-15]
 ```

File: src/tests/test_wind.f90 [hunk 1]
```diff
@@ -64,16 +63,135 @@ subroutine test_wind(ntests,npass)
...
+ nfailed(:) = 0
+ eint = sum(vxyzu(4,1:npart))
+ ekin = sqrt(sum(vxyzu(1,1:npart)**2+vxyzu(2,1:npart)**2+vxyzu(3,1:npart)**2))
+ if (vb) print '(5(1x,es22.15),i8)',xyzmh_ptmass(4,1),xyzmh_ptmass(7,1),xyzmh_ptmass(15,1),eint,ekin,npart
+ call checkval(xyzmh_ptmass(4,1),1.199987894518367E+00,epsilon(0.),nfailed(1),'sink particle mass')
+ call checkval(xyzmh_ptmass(7,1),0.,epsilon(0.),nfailed(2),'mass accreted')
+ call checkval(npart,12180,0,nfailed(3),'number of ejected particles')
+ call checkval(xyzmh_ptmass(15,1),1.591640703559762E-06,epsilon(0.),nfailed(4),'wind mass loss rate')
+ if (testcyl) then
+    call checkval(eint,3.360686893182378E+03,eps_sum,nfailed(5),'total internal energy')
```
danieljprice, line 95: It would really help here to know where these numbers come from, e.g. if we improve things how do we know if it is better? Are these just initial values, or are they from some semi-analytic solution?

src/tests/test_wind.f90
```diff
@@ -64,16 +63,135 @@ subroutine test_wind(ntests,npass)
...
+    call checkval(ekin,5.605632523862468E+01,eps_sum,nfailed(6),'total kinetic energy')
+ elseif (testkd) then
+    call checkval(eint,3.164153170427767E+03,eps_sum,nfailed(5),'total internal energy')
+    call checkval(ekin,6.101010545772693E+01,eps_sum,nfailed(6),'total kinetic energy')
+ elseif (test2) then
+    call checkval(eint,3.367417540822784E+03,eps_sum,nfailed(5),'total internal energy')
+    call checkval(ekin,5.524867074648306E+01,eps_sum,nfailed(6),'total kinetic energy')
+ else!if (test)
+    call checkval(eint,3.179016341424608E+03,eps_sum,nfailed(5),'total internal energy')
```
danieljprice, line 104: same comment, can these be calculated from first principles or are they just regression tests?

src/tests/test_wind.f90
```diff
@@ -64,16 +63,135 @@ subroutine test_wind(ntests,npass)
...
+    call allocate_array('dust_temp',dust_temp,maxTdust)
+
+    call init_testwind(2,ntests,npass,npart_old,istepfrac,dtinject)
+ !if (id==master) call write_infile('w2.in','w2.log','w2.ev','w2_00000',iwritein,iprint)
+    call integrate_wind(npart_old,istepfrac,dtinject)
+    nfailed(:) = 0
+    eint = sum(vxyzu(4,1:npart))
+    ekin = sqrt(sum(vxyzu(1,1:npart)**2+vxyzu(2,1:npart)**2+vxyzu(3,1:npart)**2))
+    if (vb) print '(5(1x,es22.15),i8)',xyzmh_ptmass(4,1),xyzmh_ptmass(7,1),xyzmh_ptmass(15,1),eint,ekin,npart
+    call checkval(xyzmh_ptmass(4,1),1.199987815414834E+00,epsilon(0.),nfailed(1),'sink particle mass')
```
danieljprice, line 127: is this m_init +  m_inj?

src/tests/test_wind.f90
```diff
@@ -97,45 +215,97 @@ subroutine test_wind(ntests,npass)
...
-      nfailed(1),'no errors in setting particle mass')
- npart_old = npart
- call inject_particles(t,0.,xyzh,vxyzu,xyzmh_ptmass,vxyz_ptmass,&
-                       npart,npart_old,npartoftype,dtinject)
- call update_injected_particles(npart_old,npart,istepfrac,nbinmax,t,dtmax,dt,dtinject)
-
- ! check 1D wind profile
- i = size(trvurho_1D(1,:))
- !print '((5(1x,es22.15)))',trvurho_1D(:,i),massoftype(igas)
- call checkval(trvurho_1D(2,i),7.058624412798283E+13,epsilon(0.),nfailed(2),'outer wind radius')
```
danieljprice, line 131: same comment with values here, can they be computed or are these just regression tests also?

src/tests/test_wind.f90
```diff
@@ -164,15 +334,6 @@ subroutine test_wind(ntests,npass)
 
  enddo
 
- !print '((3(1x,es22.15),i8))',xyzmh_ptmass(4,1),xyzmh_ptmass(7,1),xyzmh_ptmass(15,1),npart
- call checkval(xyzmh_ptmass(4,1),1.199987894518367E+00,epsilon(0.),nfailed(6),'sink particle mass')
- call checkval(xyzmh_ptmass(7,1),0.,epsilon(0.),nfailed(7),'mass accreted')
- call checkval(npart,12180,0,nfailed(8),'number of ejected particles')
- call checkval(xyzmh_ptmass(15,1),1.591640703559762E-06,epsilon(0.),nfailed(9),'wind mass loss rate')
```
danieljprice, line 171: is this prescribed somewhere?

danieljprice (commented):
thinking of future maintenance it would be helpful to know where the numbers come from :)

src/tests/test_wind.f90
lsiess: There is no analytical solution and even if it existed, it would require running the test with very high resolution and for a very log time

src/tests/test_wind.f90
lsiess: they are just regression tests

src/tests/test_wind.f90
lsiess: should be m_ini-m_inj

src/tests/test_wind.f90
lsiess: regression tests

src/tests/test_wind.f90
lsiess: xyz,h_ptmass(15) is the mass loss rate. It is defined in the .in file. Don't know why I checked this variable. It is useless.

PR #561, opened 2024-06-26
Mass transfer setup

analulujg:


build/Makefile_setups
```diff
@@ -845,6 +845,15 @@ ifeq ($(SETUP), windtunnel)
     ANALYSIS=analysis_common_envelope.f90
 endif
 
+ifeq ($(SETUP), masstransfer)
+#   Wind tunnel setup
```
danieljprice, line 849: please give a useful description as the comment here is propagated directly through to the documentation page: https://phantomsph.readthedocs.io/en/latest/user-guide/setups.html#all-setups

src/main/eos.f90
```diff
@@ -1444,13 +1444,17 @@ subroutine write_options_eos(iunit)
  endif
 
  select case(ieos)
+ case(2)
```
danieljprice, line 1447: can use case(2,12) instead of separate case statements...

src/setup/setup_masstransfer.f90
```diff
@@ -0,0 +1,215 @@
+!--------------------------------------------------------------------------!
+! The Phantom Smoothed Particle Hydrodynamics code, by Daniel Price et al. !
+! Copyright (c) 2007-2024 The Authors (see AUTHORS)                        !
+! See LICENCE file for usage and distribution conditions                   !
+! http://phantomsph.github.io/                                             !
+!--------------------------------------------------------------------------!
+module setup
+!
+! Setup of two stars or sink particles in a binary
```
danieljprice, line 9: please give a useful description here if possible (and reference to paper ideally even if it is in prep...)

danieljprice (approved):
changes look great overall, very happy to merge this with minor changes to comments

danieljprice:
build failures are because gamma is not read for a bunch of cases... seems to be caused by the changes to the case statement in read_options_eos ?

PR #577, opened 2024-07-22
Add a new star formation prescription and HII region expansion stellar feedback (Embedded clusters simulation Framework)

Yrisch:
Type of PR: 
New physics

Description:

**On the star formation prescription** : activated with icreate_sinks == 2. This new star formation process consider sinks as stellar cores that could contain few stellar seeds. Cores will accrete mass using standard accretion method used with sinks. At a given fixed time this accretion stop and the sink (core) is converted into stars. These stars are treated as point masses without accretion radius. 
Only two parameters are currently free : the maximum accretion time `tmax_acc` and the creation time of the seeds `tseeds`
The prescription is based on two main functions : `ptmass_createseeds` that initiate virtual ptmasses (with negative mass) after a formation time (before this time the sink is considered as a dense clump of gas). It is mandatory to count the right number of seeds within a core if we have multiple merging event before accretion stop.
It also uses a linked list `linklist_ptmass` to track the number of seeds within each sinks. Each sinks after seeds creation has a sub linked list of those ones. This sub linked list can merge together if a merging event occurs.
When the sink (core) reach the maximum accretion time the seeds within it are converted into stars (ptmass with tiny accretion radius)  using `ptmass_createstars`, and they are placed into the simulation using a Plummer distribution (Aarseth 1974) inside the accretion radius of the parent sink. The momentum of the parent sink is also conserved. The parent is killed after that.

**On the HII stellar feedback** : Simple scheme to reproduce the HII region expansion aroud massive stars (> 8 MSun) in star forming region. This method uses the approach developed in Hopkins(2012) and Fujii+ (2021). Each massive stars are associated with a ionizing rate computed using table fitting (Fujii + 2021). It is then need to list the neighboring gas particles in order. `getneigh_pos` gives all the neighbors around a position then `Knn_func` sort the particles index in an ascending order. After that the main loop iterate on this list and compute for each particle the ionization rate needed to fully ionize the particle. This rate is then subtracted from the star ionizing rate. The loop continues until this rate is equal to zero. each particle ionized is flagged to true using `isionised` array.
Each of them are then set to 10⁴ K in a new EoS. 
`ieos == 21`  Isothermal in cold (10K) and warm(10⁴ K) media
`ieos == 22`  Isothermal in ionized regions(10⁴K) but adiabatic+cooling in cold medium (need to investigate this one)

Other few adds : 
-New timers for HII region and subgroups
-Modification of the cluster setup
-Fix some issues

Testing:
New unit test for HII regions : Compare the initial Strömgren radius from the prescription and the theoritical formula
New unit test for icreate_sinks == 2

Did you run the bots? yes


Yrisch:
Weird bug with build test... It first builds correctly and crashes at the second build to test the setup for no reasons...

Yrisch:
Note : This new star formation scheme needs to store `linklist_ptmass` in a full dump to restart a simulation... It adds an 1D integer array to the `ptmass` block. I spotted an issue to open this new full dump with **splash** . In the splash read routine, we need now to skip all integer arrays as done with the gas block.

src/main/writeheader.F90
```diff
@@ -182,15 +182,19 @@ subroutine write_header(icall,infile,evfile,logfile,dumpfile,ntot)
        if (drag_implicit) then
           write(iprint,"(1x,a)") 'Two-fluid dust implicit scheme is ON'
        else
-          write(iprint,"(1x,a)") 'Two-fluid dust explicit scheme is ON'
+          write(iprint,"(1x,a)") 'Two-fluid dust explicit scheme is OFF'
```
danieljprice, line 185: this should be ON (the word change here is implicit->explicit)

src/setup/set_orbit.f90
```diff
@@ -44,6 +44,51 @@ module setorbit
...
+!
+! :References: None
+!
+! :Owner: Daniel Price
+!
+! :Runtime parameters: None
+!
+! :Dependencies: infile_utils, physcon, setbinary, setflyby, units
+!
+
```
danieljprice, line 61: header comments seem to have multiplied here...

src/main/evolve.F90
```diff
@@ -276,14 +285,57 @@ subroutine evol(infile,logfile,evfile,dumpfile,flag)
...
+       !
+       if (ipart_createstars /= 0) then
+          call ptmass_create_stars(nptmass,ipart_createstars,xyzmh_ptmass,vxyz_ptmass,fxyz_ptmass,fxyz_ptmass_sinksink, &
+                               linklist_ptmass,time)
+       endif
+    endif
+
+    if (iH2R > 0 .and. id==master) then
+       istepHII = 1
+#ifdef IND_TIMESTEPS
```
danieljprice, line 309: please use if (ind_timesteps) here to avoid the ifdef

danieljprice:
> Note : This new star formation scheme needs to store `linklist_ptmass` in a full dump to restart a simulation... It adds an 1D integer array to the `ptmass` block. I spotted an issue to open this new full dump with **splash** . In the splash read routine, we need now to skip all integer arrays as done with the gas block.

-> should raise an issue on the splash GitHub

PR #583, opened 2024-08-08
Slow down method and few fixes on the new star formation prescription

Yrisch:
Type of PR: 
new physics

Description:
Slow down method during subgroups integration. This method slows down the inner motion of weakly perturbed binaries in subgroups to relax constraints on the time step. It is still possible to follow secular evolution of this binaries without integrating their real fast motion. We loose the orbital phase but the computational gain is significant. See Mikkola and Aarseth (1996) and Wang et al. (2020) for more details.

Testing:
This new feature has been tested using the setup hierarchical reproducing tests done in Wang et al. (2020)... 

Did you run the bots? yes


danieljprice:
this looks ok to merge. Main question is, how would you know if this is broken in the code?

What we really need here to keep this from being broken is an automated unit test, for example could we repeat some of the tests in Wang et al. (2020) as part of the test suite...?

Yrisch:
Yes, you right. I will work on it !

Yrisch:
After a lot of fixes and fine tuning, the test should be almost ready. I checked manually the results and it's perfect. I'm finishing automatic checks and It should be good to go !

Yrisch:
I think everything is fine now... The only failing seems to be a failed download...

danieljprice (approved):

PR #591, opened 2024-09-04
Update to analysis_common_envelope

themikelau:
It is time to update this analysis module.

Changes to analysis_common_envelope:
- Option to output extra quantities to ".divv" files replaced with a proper implementation, where an arbitrary number of extra quantities can be written to ".extras" files, which can be read by splash (see https://github.com/danieljprice/splash/pull/87)
- Most important options are made compatible with simulations run with radiation.

Changes to other parts of Phantom:
- Function`entropy` in `eos` module: Add option to specify radiation temperature directly, which is useful if using radiation and Trad is stored.
- `radiation_utils`: Redefine `radE_from_Trad` (output is energy per unit volume) to `radxi_from_Trad` (output is energy per unit mass), the latter being more useful. Likewise, the inverse function is also changed from `Trad_from_radE` to `Trad_from_rhoxi`. The rest of the code has been adjusted to be consistent with these new definitions.
- `eos_idealplusrad`: Add functions to calculate gas energy and radiation energy separately.


themikelau:
The test suite caught a bug with the cons2prim test. There was an issue because I have lumped some changes to entropy calculation in this pull request. Namely, I have ensured that any temperature solving uses the gas constant, `Rg`, instead of `kb_on_mh`. Mixing the two constants causes a discrepancy great enough to fail the cons2prim tests. But I forgot to change this in the temperature solving in `get_p_from_rho_s`. This has been fixed now. Can @Fitz-Hu check if these changes are correct?

danieljprice:
See comment on splash PR, would be better to use .cols file which is an existing splash feature

danieljprice:
Also the cons2prim stuff should be pretty well unit tested so happy to approve that

PR #596, opened 2024-10-17
(set-binary) update npartoftype array when particle types are set according to a specific star in a binary

themikelau:
Type of PR: 
Bug fix (found by @msha0023)

Description:
When setting up two gas-particle stars in a binary using `SETUP=binary` with `relax_star`, the second star is not relaxed because of an error raised by `checksetup` called in `relax_star:
```
rstar =    1.0000000000000000       mstar =   0.99935806369259794       tdyn =    1.1110774122591152     
 n(via iphase)=         100           0           0         100           0           0           0           0
 npartoftype  =         200           0           0           0           0           0           0           0
 ERROR: sum of types in iphase is not equal to npartoftype

 ERROR! relax_star: cannot relax star because particle setup contains errors
```
The cause of this is that the `npartoftype` array was not updated when particles from star 1 were set to the type `itype+ioffset`, causing a mismatch with `iphase`. I have now ensured `npartoftype` is updated whenever particle types are changed.

### DIFF
--- a/src/setup/set_star.f90
+++ b/src/setup/set_star.f90
@@ -802,17 +802,10 @@ subroutine read_options_star(star,need_iso,ieos,polyk,db,nerr,label)
 
  select case(star%iprofile)
  case(imesa)
-    ! core softening options
-    call read_inopt(star%isinkcore,'isinkcore'//trim(c),db,errcount=nerr)
-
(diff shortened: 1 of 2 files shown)


danieljprice:
thanks both

PR #610, opened 2024-12-17
Radiative cooling approximation

alisonkyoung1:
Type of PR: 
new physics

Description:
Added the Modified Lombardi cooling approximation as icooling = 9. I have added a new equation of state (ieos=24) which is used by the cooling. There is also an additional setup parameter to use for modelling self-gravitating discs with this method that uses the luminosity of the central star.

This version does not include FLD (via the Forgan+ 2009 approach) because I didn't want to duplicate code.

Testing:
The code reproduces the expected temperature & density trend for the collapse of a uniform sphere. EOS test included with the other EOS tests.

Did you run the bots? yes- but it now says I'm the owner of e.g. eos.f90


src/main/step_leapfrog.F90
```diff
@@ -420,15 +431,12 @@ subroutine step(npart,nactive,t,dtsph,dtextforce,dtnew)
...
     dt_too_small = .false.
-
     call derivs(1,npart,nactive,xyzh,vpred,fxyzu,fext,divcurlv,&
                 divcurlB,Bpred,dBevol,radpred,drad,radprop,dustproppred,ddustprop,&
                 dustpred,ddustevol,filfacpred,dustfrac,eos_vars,timei,dtsph,dtnew,&
                 ppred,dens,metrics,apr_level)
-
     if (do_radiation .and. implicit_radiation) then
        rad = radpred
-       vxyzu(4,1:npart) = vpred(4,1:npart)
```
danieljprice, line 431: this line seems dangerous to remove as may mess up implicit FLD... should not affect icooling=9 anyway

src/main/force.F90
```diff
@@ -1189,6 +1193,12 @@ subroutine compute_forces(i,iamgasi,iamdusti,xpartveci,hi,hi1,hi21,hi41,gradhi,g
  fgravxi = 0.
  fgravyi = 0.
  fgravzi = 0.
+ if (icooling == 9) then
+    gradP_cool(i) = 0d0
```
danieljprice, line 1197: house style here is 0.  not 0d0 as we automatically promote real->double

src/main/force.F90
```diff
@@ -1030,6 +1033,7 @@ subroutine compute_forces(i,iamgasi,iamdusti,xpartveci,hi,hi1,hi21,hi41,gradhi,g
  real    :: dlorentzv,lorentzj,lorentzi_star,lorentzj_star,projbigvi,projbigvj
  real    :: bigvj(1:3),velj(3),metricj(0:3,0:3,2),projbigvstari,projbigvstarj
  real    :: radPj,fgravxi,fgravyi,fgravzi
+ real    :: gradpx,gradpy,gradpz,gradP_cooli=0d0,gradP_coolj=0d0
```
danieljprice, line 1036: house style is 0. not 0d0, also suggest to initialise these in the routine itself rather than on the declaration line

src/main/eos.f90
```diff
@@ -662,6 +685,7 @@ subroutine get_TempPresCs(eos_type,xyzi,vxyzui,rhoi,tempi,presi,spsoundi,gammai,
  endif
 
  if (maxvxyzu==4) then
+    if (vxyzui(4) < 0d0) print *, "ui NEGATIVE in eos"
```
danieljprice, line 688: please replace this with a "call warning('eos','ui negative in eos')" so that the warnings buffer correctly

docs/physics/eos-list.rst
```diff
@@ -158,3 +158,9 @@
    |           | reference density :math:`\rho_0`, e.g. 2.7 g/cm^3 for basalt                                                             |
    |           |                                                                                                                          |
    +-----------+--------------------------------------------------------------------------------------------------------------------------+
+   | 24        | **Tabulated EoS of Stamatellos et al. 2007 (includes opacities)**                                                        |
```
danieljprice, line 161: this doc file is written automatically from the code in eos.f90, suggest to move/duplicate the comments there so they don't get removed by the script (cd docs/scripts; ./update_docs_from_code.sh)

src/main/cooling_radapprox.f90
```diff
@@ -0,0 +1,251 @@
...
+
+ coldensi = huge(coldensi)
+ kappaBari = 0d0
+ kappaParti = 0d0
+ Teqi = huge(Teqi)
+ tthermi = huge(tthermi)
+ opaci = epsilon(opaci)
+ if (abs(ui) < epsilon(ui)) print *, "ui zero", i
+
+ if (isink_star > 0) then
```
danieljprice, line 132: these should be **2 (integer) not **2d0 (double) for efficiency

src/main/eos.f90
```diff
@@ -480,6 +482,20 @@ subroutine equationofstate(eos_type,ponrhoi,spsoundi,rhoi,xi,yi,zi,tempi,eni,gam
     ponrhoi  = real(cgspresi / (unit_pressure * rhoi))
     spsoundi = real(cgsspsoundi / unit_velocity)
     !  tempi    = 0. !temperaturei
+case (24)
+!--interpolate tabulated eos from Stamatellos+(2007). For use with icooling=9
```
danieljprice, line 486: insert the comments here that you want to appear in the documentation

src/main/eos_stamatellos.f90
```diff
@@ -0,0 +1,276 @@
...
+ use allocutils, only:allocate_array
+
+ print *, "Allocating cooling arrays for maxp=",maxp
+ call allocate_array('gradP_cool',gradP_cool,maxp)
+ call allocate_array('ttherm_store',ttherm_store,maxp)
+ call allocate_array('ueqi_store',ueqi_store,maxp)
+ call allocate_array('opac_store',opac_store,maxp)
+ call allocate_array('duSPH',duSPH,maxp)
+
+ gradP_cool(:) = 0d0
```
danieljprice, line 45: not very important but house style is 0. not 0d0

src/main/eos_stamatellos.f90
```diff
@@ -0,0 +1,276 @@
...
+
+ integer i,j
+ real m,c
+ real kbar1,kbar2
+ real kappa1,kappa2
+ real Tpart1,Tpart2
+ real gmw1,gmw2
+ real ui_, rhoi_,rhomin,umin
+
+ rhomin = OPTABLE(1,1,1)
```
danieljprice, line 117: suggest OPTABLE->optable to reduce shouting

src/main/force.F90
```diff
@@ -2974,30 +3000,33 @@ subroutine finish_cell_and_store_results(icall,cell,fxyzu,xyzh,vxyzu,poten,dt,dv
...
 
        if (mhd) then
           !
-          ! sum returns d(B/rho)/dt, just what we want!
+          ! sum returns d(b/rho)/dt, just what we want!
           !
-          dBevol(1,i) = fsum(idBevolxi)
-          dBevol(2,i) = fsum(idBevolyi)
-          dBevol(3,i) = fsum(idBevolzi)
+          dbevol(1,i) = fsum(idbevolxi)
```
danieljprice, line 3014: please revert case to dBevol to follow rest of code, also in comment below, looks like a search/replace issue

src/main/force.F90
```diff
@@ -3018,6 +3047,7 @@ subroutine finish_cell_and_store_results(icall,cell,fxyzu,xyzh,vxyzu,poten,dt,dv
        ! cooling timestep dt < fac*u/(du/dt)
        if (maxvxyzu >= 4 .and. .not. gr) then ! not with gr which uses entropy
           if (eni + dtc*fxyzu(4,i) < epsilon(0.) .and. eni > epsilon(0.)) dtcool = C_cool*abs(eni/fxyzu(4,i))
+          if (dtcool < epsilon(0.) .or. isnan(dtcool)) print *, "dtcool=zero or NaN in force.F90", dtcool
```
danieljprice, line 3050: suggest call fatal or call warning here rather than a print statement

danieljprice (approved):
looks good overall, thanks Alison! A few minor things, only one I am worried about is the one where vxyzu=vpred has been commented out, I think this is needed for the implicit FLD... please check

src/main/force.F90
alisonkyoung1: (Note to self) I think this statement is no longer needed...

src/main/step_leapfrog.F90
alisonkyoung1: Sorry, definitely a mistake!

alisonkyoung1:
I've fixed the above issues and it should be good to go

danieljprice:
thanks Alison for the contributions here!

PR #615, opened 2025-01-29
fix windtunnel injection so it works for arbitrary wind parameters

danieljprice:
Type of PR: 
modification to existing code

Description:
This p-r fixes the wind tunnel injection and setup options so that they are independent of the desire to add a "star" to the simulation.
- add add_star flag to the .setup file instead of a windonly flag in the injection module
- use the hold_star flag in the inject options to decide whether to adjust velocities inside the star. If this option is off, we do not ask for Mstar or Rstar in the injection options
- ask for Mach number rather than pressure in the injection parameters, as much simpler to understand (and dimensionless). 
- default parameters are set to give the same pres_inf as was previously hardwired (i.e. Mach 13.86)
- take particle mass directly as the input parameter for the mass resolution if no star is added to the setup
- wind_length, wind_radius and wind_injection_x parameters are now taken in code units rather than being in units of Rstar (since Rstar is undefined if no star is added)

Testing:
```~/phantom/scripts/writemake.sh windtunnel > Makefile
make setup; make
./phantomsetup wind
./phantom wind
```

Did you run the bots? no

### DIFF
--- a/src/main/inject_windtunnel.f90
+++ b/src/main/inject_windtunnel.f90
@@ -32,27 +32,26 @@ module inject
  character(len=*), parameter, public :: inject_type = 'windtunnel'
 
  public :: init_inject,inject_particles,write_options_inject,read_options_inject,&
-           set_default_options_inject,update_injected_par,windonly
+           set_default_options_inject,update_injected_par
 !
 !--runtime settings for this module
 !
-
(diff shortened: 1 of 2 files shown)


themikelau (approved):
Thanks!

PR #620, opened 2025-02-13
Sinks in the kd tree 

Yrisch:
Type of PR: 
modification to existing code

Description:
Add an option to construct the kd tree using SPH and sink particles. It is then possible to calculate sink-gas and sink-sink long-range interactions using the main force routine.
This can be useful when the number of sinks inside a simulation is high. In that case, the direct sink-gas direct $\mathcal{O}(N_\text{sink}N_\text{SPH})$ interaction starts to be the main bottleneck. The new method can bring back the complexity of this interaction to $\mathcal{O}(N_\mathrm{sink}\mathrm{log}(N_\mathrm{gas}) + N_\mathrm{gas}\mathrm{log}(N_\mathrm{sink}))$ at the cost of 2nd order multipole accuracy. To use with care!
The implementation uses a precompilation flag that will change the `maxp` value to  `maxp = maxp + maxptmass`. This ensures to have enough space in `xyzh_soa` and `iphase_soa` to push sink positions, masses and the new `isink` phase. The only modification in the tree construction is a loop on point masses in `construct_root_node` to include them in arrays and node CoM/size calculation. Despite that, the local tree is built the same way it was before. 

This implementation is also MPI-compatible. To construct the global tree, The ptmass indices are spread over every mpi process during the load balancing. That way, each process can build its local tree only with ptmass in its domain.

Few conditions in other modules are necessary to avoid using sinks in neighbour loops. The main force loop has been modified to identify sink particles. Sink-gas short-range interactions are softened the same way as the direct method. short sink-sink interactions are skipped to be calculated within the substep routine.

Finally, during one step, sink velocities need to be kicked with these new forces during the upper-level integration like SPH particles. To agree with the Individual timestep scheme, sinks are always considered active (smallest bin) for now. 

Testing:
Grav and ptmass tests applied with SINKTREE=yes

Did you run the bots? no


src/tests/test_ptmass.f90
```diff
@@ -231,6 +232,9 @@ subroutine test_binary(ntests,npass,string)
        if (use_fourthorder) then
           if (id==master) write(*,"(/,a)") '--> skipping integration of binary orbit with oblateness with FSI'
           cycle binary_tests
+       elseif (use_sinktree) then
+          if (id==master) write(*,"(/,a)") '--> skipping integration of binary orbit with oblateness with Sinktree'
```
danieljprice, line 236: can you explain why these tests do not work if the sinks are in the tree?

danieljprice:
all looks good from my side... only question is why the binary tests do not work if sink particles are in the tree --- surely they should still pass in this case?

src/tests/test_ptmass.f90
Yrisch: Yes you right... oblateness should not be a problem if sink-sink interactions stay within the kd-tree short range. This exception can be removed.
However the sink-gas case can not work currently with sinks in the tree.

Yrisch:
I can't reproduce the issue in the last check all the time.  I also tried to do it on the master branch and it seems broken too. I don't understand... There is probably a race condition somewhere.

danieljprice:
yes, the accretion test seems to fail sporadically, this is unrelated to your pull request, seems to work if re-run

PR #624, opened 2025-02-21
(grtde) Add option to compute initial velocity of a star using GR initial conditions

maguilarf:
Type of PR: 
Modification to existing code

Description:
In the .setup file for setting up stars with grtde, users can now choose to calculate the initial velocity of a star in General Relativity (GR) for parabolic orbits around non-spinning black holes.
- Added a boolean parameter use_gr_ic in the .setup file to enable the calculation of GR initial conditions.
- The initial velocity of the star is computed in Boyer-Lindquist coordinates in setup_grtde.f90 (see, e.g., Appendix in Tejeda et al. 2017, MNRAS, 469, 4483–4503).
- The total specific energy, calculated within GR, is used to numerically determine the initial velocity via the Gradient Descent method.

### DIFF
--- a/src/setup/setup_grtde.f90
+++ b/src/setup/setup_grtde.f90
@@ -52,6 +52,7 @@ module setup
  integer :: dumpsperorbit,nstar
  logical :: relax,write_profile
  logical :: provide_params
+ logical :: use_gr_ic
  integer, parameter :: max_stars = 2
  type(star_t)  :: star(max_stars)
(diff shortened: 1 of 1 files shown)


danieljprice:
looks great - very helpful. It may be possible to do this in a more general way using the metric routines rather than hardwiring Boyer-Lindquist coordinates, but happy to merge this as it's already useful. Thanks for the contributions!

Issue #635, opened 2025-04-16
seg fault related to sinks in tree


danieljprice:
(reported by Vasu Prasad)

restarting from a gas/dust simulation dump gives a horrible seg fault, seems to be related to gas particles being misidentified as sink particles in the tree build routine

unclear why this is occurring, seems to be related to recent merges of sink-particles-in-the-tree functionality. Under active investigation...

danieljprice:
now fixed on master

PR #637, opened 2025-04-17
(#635) fix seg fault caused by shuffling of particle types by 1

danieljprice:
Type of PR: 
Bug fix

Description:
Fix for #635 reported by Vasu Prasad. The underlying issue was breaking of backwards compatibility with previous runs using dust particles because the dust particle type had been shuffled by +1 to make room for the isink particle type.

A better solution is to make the isink type needed to identify sink particles in the tree equal to 0. This works because it is also the "unknown" particle type, does not correspond to an entry in the massoftype array, and does not need a sign when referenced as iphase because sink particles cannot be active/inactive.

Testing:

Using dump file standard_run_00122 supplied by Vasu. This file has 10^6 gas particles (type 1) and 10^5 dust particles of type 7. The seg fault occurred because the type 7 particles read from the dump were incorrectly identified as sink particles during the tree build.
```
./phantom standard_run.in
```

Did you run the bots? no

### DIFF
--- a/src/main/checksetup.f90
+++ b/src/main/checksetup.f90
@@ -251,6 +251,11 @@ subroutine check_setup(nerror,nwarn,restart)
 !--check that mass of each type has been set
 !
  do itype=1,maxtypes
+    if (isnan(massoftype(itype))) then
+       print*,'WARNING: massoftype = NaN for '//trim(labeltype(itype))//' particles'
+       nwarn = nwarn + 1
+       massoftype(itype) = 0.
+    endif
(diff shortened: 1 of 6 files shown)


Yrisch:
Hi,
My bad! I didn't think about the backward compatibility with previous runs... I agree with your fix, it was my first idea. But `isink = 8` looked cleaner at that moment. I'm looking at potential side effects with this new definition.

Yrisch:
I verified the previous failed tests.  `get_partinfo` tagged all sink particles as inactive. It can be solved by allowing particles with `itype >=0` to be active.

danieljprice:
Thanks Yann for the quick fix to test failure!

PR #640, opened 2025-04-21
Eccentric discs

eragusa:
Type of PR: 
modification to existing code to enable initialisation of eccentric discs + the possibility to initialise both eccentricity profile and sigma profile from a tabulated file.

Description:
Setup_disc and set_disc are modified to initialise eccentric discs with a prescribed eccentricity profile. Phantomsetup now interactively asks if the user wants to initialise an eccentric disc. If the user answers yes, additional lines are included in the .setup file for specifying the desired profile. 

The user can prescribe to use a constant eccentricity profile, a power law, or to be read from a two column file [a, e] ("a" being semi-major axis and "e" the eccentricity value) read from the file etc_grid.dat. Similarly, now phantomsetup can initialise the density profile reading the profile from a file [a, Sigma], sigma_grid.dat (if the disc is eccentric, what is initialised is Sigma(a)).

A script for generating ecc_grid.dat and sigma_grid.dat can be found in the scripts folder (generate_eccsigma_grid.py). It acts as a template, you should edit the functions in the script to whatever profile you want and make sure you run the script before running phantom setup.

I included the script sigma_ecc_analysis.py that reads the dump and plots e(a) and Sigma(a) profiles, you can use it either to plot the e(a) and Sigma(a) profiles in your simulations or to make sure the setup initialises the disc as you like it. I tested it extensively, if you find instances in which it does not do what it should, please let me know.

Testing:
The setup produces the correct sigma and eccentricity profile in the first dump compared to sigma_ecc_analysis.py namedump_00000, so to me the setup does what I would like it to do. 

Did you run the bots? yes
Bots are mostly fine (170/188), the failing ones are mysterious to me, though: they appear to be a few cases of SETUP=test2, and systematically the test "testing integration of disc around eccentric binary with Forward Symplectic Integrator" SETUP=testcyl, where linear momentum is not conserved. I spent a substantial amount of time trying to trace the origin of these failures without success. It is puzzling, because when the setup is initialised to default values as during tests, eccentricity is set to zero and the equations all reduce to the old circular setup initialisation (they really do, I tested this veryfing that during those tests the semi-major axis=R when ecc=0; and Keplerian velocity vphi_k equals sort(GM/R)). So from the perspective of the tests, set_disc and setup_disc were not affected, but apparently they were in some unintelligible way.
I find also curious that the same tests involving set_disc in some other SETUPs performing them work fine, so I am not sure what is going on there. 


src/setup/set_disc.F90
```diff
@@ -801,21 +928,24 @@ subroutine adjust_centre_of_mass(xyzh,vxyzu,particle_mass,i1,i2,x0,v0)
...
-    if (i_belong_i4(i)) then
-       ipart = ipart + 1
-       xcentreofmass = xcentreofmass + particle_mass*xyzh(1:3,ipart)
-       vcentreofmass = vcentreofmass + particle_mass*vxyzu(1:3,ipart)
-       totmass = totmass + particle_mass
-    endif
- enddo
+     if (i_belong_i4(i)) then
+        ipart = ipart + 1
+        if((e_0 == 0.) .and. (ecc_profile .ne. 4)) then
```
danieljprice, line 933: this looks dangerous, one should never check equality for reals, instead use abs(e_0) < tiny(e_0)

src/utils/interpolate_grid.f90
```diff
@@ -0,0 +1,114 @@
+module interpolate_grid
```
danieljprice, line 1: if possible it would be better to put these routines in src/main/utils_tables.f90, where you will find at least two other 1D interpolation routines :)

src/utils/load_from_file.f90
```diff
@@ -0,0 +1,181 @@
...
+    open(unit=iunit,file=namefile,status='replace',action='write')
+    do i=1,size(arraytowrite1(:))
+       write(iunit,*) arraytowrite1(i),arraytowrite2(i)
+    enddo
+    close(unit=iunit)
+
+ end subroutine write_in_file_1dx2    
+ 
+
+ integer function number_of_columns(s,nhead)
```
danieljprice, line 122: see also routines in src/main/utils_filenames.f90 with similar functionality

src/setup/set_disc.F90
```diff
@@ -567,24 +641,45 @@ subroutine set_disc_positions(npart_tot,npart_start_count,do_mixture,R_ref,R_in,
  ipart = npart_start_count - 1
 
  !--loop over particles
- do i=npart_start_count,npart_tot,2
-    if (id==master .and. mod(i,max(npart_tot/10,10))==0 .and. verbose) print*,i
+ do i=npart_start_count,npart_tot,1
```
danieljprice, line 644: I think one of the test failures is because you removed my "symmetric disc" feature that keeps the centre of mass exactly at the origin by placing particles in pairs around the origin. I think it would be useful to keep this feature for the (typical) case where e=0., can perhaps have a simple integer flag n_to_place = 1 for e > 0 and n_to_place = 2 for the circular case

src/setup/set_disc.F90
```diff
@@ -627,26 +722,46 @@ subroutine set_disc_positions(npart_tot,npart_start_count,do_mixture,R_ref,R_in,
...
-       xyzh(1,ipart) = R*cos(phi)
-       xyzh(2,ipart) = R*sin(phi)
+       xyzh(1,ipart) = R_ecc*cos(phi)
+       xyzh(2,ipart) = R_ecc*sin(phi)
        xyzh(3,ipart) = zi
        xyzh(4,ipart) = hpart
        !--set particle type
        call set_particle_type(ipart,itype)
-    endif
-    if (i_belong_i4(i+1) .and. i+1 <= npart_tot) then
```
danieljprice, line 640: could keep this with an ".and. n_to_place == 2"

src/setup/set_disc.F90
eragusa: I fixed this and other instances where I checked equality or inequality with reals.

src/setup/set_disc.F90
eragusa: I guess your are right. I restored your feature as you suggested. It is testing, let's see if that was the issue.

src/setup/set_disc.F90
eragusa: As above, I introduced the n_to_place==2.

src/utils/load_from_file.f90
eragusa: They are currently unused, I moved them in utils_filesnames.f90 and commented them for the moment to avoid warnings during compilation. I guess I can remove them if they are redundant.

eragusa:
I fixed all the failures I was getting, but now it appears "MPI test (SETUP=test2, targets=, DEBUG=yes, OPENMP=yes, ntasks=1)" fails, but not for ntasks=2. I am not sure how to solve this. The same test on the last commit yesterday did not fail. Since then I only changed a few dependencies In the Makefile to solve some build issues I got that were the only fails I was getting. Unless the test itself changed, I have no idea why that is the case. All tests on my fork (last pull from your master 3 weeks ago) are successful. I'll try to pull your master branch and see if that fixes it.

PR #642, opened 2025-04-24
Radiation tests

themikelau:
**Radiation tests in [Lau et al. (2025)](https://ui.adsabs.harvard.edu/abs/2025arXiv250320506L/abstract)**

**Gas heating/cooling tests**
- Forty logarithmic time steps for the implicit solver (see plot) instead of `dt = 0.05*t`.
- ~~The problem is that with logarithmic time steps, the time steps closer to the end time are larger compared to before, leading to a greater discrepancy with the explicit solution. Previously, the test tolerance was 2e-15. Now, the implicit and explicit solvers only agree to within 10^-7 in the last value in erad. This is currently causing the test failures. We could just lower the tolerance to 10^-7... What do you think, @danieljprice ?~~ **FIXED**

Screenshot 2025-04-24 at 16 13 27

**Radiative shocks**
Screenshot 2025-04-25 at 17 40 42

**Radiation pulse with implicit radiation solver**
- Does not work with explicit solver yet.
- The radiation pulse test requires boundary particles held at fixed radiation energy densities. For boundary particles, we set `mask(i) = .false.` so that they are not updated in the implicit iterations. 

Screenshot 2025-04-25 at 17 29 24

**Shock setup**
- Added radiation pulse case to `setup_shock`.
- Add radiation energy to the `leftstate` and `rightstate` vectors to set up a discontinuous jump in xi.
- Allow setting arbitrary position for discontinuity through the parameter `xshock`.
- Radiation pulse requires units of (M,L,T) = (g, cm, 10^-4 s) when using very short time steps (e.g., to compare with the explicit timestep). Otherwise, numbers become smaller than epsilon and crash the implicit solver. At the moment, the implementation of special units for the radiation pulse is ugly but there is no straightforward way around it.
- New flag `set_radiation_and_gas_temperature_equal` to indicate whether to assume thermodynamic equilibrium, which automatically sets the left and right states for xi from u, rho. This flag is set to false for the radiation pulse case.

**Other changes:**
- Default accuracy tolerance for implicit radiation solver decreased from 10^-4 to 10^-6 following Lau et al. (2025).
- `set_units_interactive` now actually calls `set_units`.


src/tests/test_radiation.f90
```diff
@@ -102,11 +102,12 @@ subroutine test_exchange_terms(ntests,npass,use_implicit)
  real :: pmassi,rhozero,totmass
  integer, intent(inout) :: ntests,npass
  logical, intent(in)    :: use_implicit
- real :: dt,t,physrho,rhoi,maxt,laste
- integer :: i,nerr(1),ndiff(1),ncheck,ierrmax,ierr,itest
+ real :: dt,t,physrho,rhoi,maxt,laste,dt0
+ integer :: i,nerr(1),ndiff(1),ncheck,ierrmax,ierr,itest,N_implicit_steps
  integer(kind=8) :: nptot
- logical, parameter :: write_output = .false.
+ logical, parameter :: write_output = .true.
```
themikelau, line 108: Need to undo this

themikelau:
@danieljprice Happy to merge this if you think the changes are okay.

Issue #487, opened 2023-12-15
ionisation schemes + equation of state


danieljprice:
currently we have a variety of different modules designed to capture ionisation and recombination, including the following that assume ionisation/recombination equilibrium:

1. the "gasradrec" equation of state
- assumes ionisation equilibrium
- works with any composition
- does H and He ionisation but does not currently do H2 formation/dissociation 
- adapts mu so that u is conserved but does not update gamma (since it would only change with H2 formation)

2. the tabulated MESA equation of state
- assumes ionisation equilibrium
- adapts mu/gamma so that u is conserved
- requires fixed composition (X,Y,Z)
- probably does not include H2

4. the non-tabulated Helmholtz equation of state
- assumes ionisation equilibrium
- the reverse solver for T from u is not implemented
- composition could be local in principle
- assumes full ionisation so valid only for high temperatures

5. NICIL library for the non-ideal MHD coefficients
- includes cosmic ray and thermal ionisation of H and He (and maybe also H2 formation/destruction?)
- assumes chemical equilibrium
- currently not used to change the equation of state, though could be in principle?

6. ieos=5 which
- does H2 formation/destruction but not ionisation
- adapts mu/gamma so that u is conserved

7. the H2 chemistry module (h2chem.f90) 
- does not assume ionisation or chemical equilibrium
- tracks formation/destruction of H2 and ionisation of H but not He
- updates mu and could be used to update gamma in principle

8. cv_table for the flux limited diffusion radiation hydro
- assumes fixed composition (X,Y,Z)
- includes H/He ionisation and also H2 formation/destruction
- tabulated eos, but we produce the table so can be extended in principle

The working plan is to try to consolidate these efforts into a few "master" prescriptions. 

Suggested steps are:

- extend gasradrec to deal with H2 formation/destruction and variable gamma
- get rid of ieos=5 and merge into gasradrec
- keep MESA as reference but implement unit tests that check that results are the same in the overlap region
- hopefully this means we can remove the cv_table (item 8), this needs to be checked for speed in the implicit radiation solver
- output and store n_e, mu and gamma as part of the equation of state calculation, for use in MCFOST atomic line transfer

PR #650, opened 2025-05-02
Gas+rad+rec EoS: Include changes in effective gamma due to H2 vibrational and rotational modes

### LINKED ISSUE #654: Radiation energy not set up in SETUP=raddisc
Issue reported by Hans Lee:
When using the raddisc setup, particles have zero radiation energy (this produces a warning). The correct behaviour is to set the radiation energy assuming thermodynamic equilibrium (Tgas=Trad). This is fixed when uncommenting the call to `set_radiation_and_gas_temperature_equal` in `setup_disc.f90`. It would be good to know if there was a specific reason for commenting out that line in the first place that I am missing.

themikelau:
Main changes:
- `ionization_mod`: Implement Ryo's fits to the molecular component of c_v following Black & Bodenheimer (1975).
- `equationofstate` to take `rad(ixi,i)` as argument as it is needed for sound speed calculation when Tgas ≠ Trad. 
  - However, this has not yet been fully implemented as `cons2prim_everything` has not been updated to pass `rad(ixi,i)` to the EoS call.
- Many subroutines in `eos_gasradrec` (and also `calc_temp_and_ene` in `eos`) now take `do_radiation` as an input. When `do_radiation=.true.` do not include radiation contributions to the total pressure and internal energy because they are treated separately.
- Add the eos_gasradrec case to `get_cv` with different treatments depending on `do_radiation`.
  - When `do_radiation=.true.`, return `cv=u/T`.
  - When `do_radiation=.false.`, return `cv=(u-u_rad)/T`. This change allows implicit radiation hydro to work with the gas+rad+rec EoS.
- `set_star_thermalenergy`: Add case for gas+rad+rec EoS with radiation and clean up. 
- `radiation_utils`: Get rid of some subroutines that are redundant or not very useful: `T_from_Etot`, `get_1overmu`.

Unit testing:
- Add check for `cv` in forward-backward consistency test for the gas+rad+rec EoS.
- Increase tolerance for gas+rad+rec EoS forward-backward consistency test from 1e-14 to 2e-14.
- Increase tolerance for ideal gas+rad EoS forward-backward consistency test from 1e-15 to 2e-15.
- Adjust tested parameter range in EoS forward-backward test from logQ = [-10,-2], logT = [1,8] to logQ = [-10,2], logT = [2,8].
- Add subroutine in `test_eos` to print data table of molecular c_v vs. temperature.

Other changes
- Get rid of errors that block usage of `ieos=20` with radiation.
- Remove subroutine `calc_rec_ene` from `eos` module as it is replaced by Ryo's much more advanced `ionization_mod`.
- Bug fix: Correct logic for when to write `gmw` into infile.
- Convert scalar and vector versions of `rapid_tanh`, `rapid_dtanh` into elemental functions.
- `analysis_common_envelope`: Clean up subroutine `eos_surfaces` and add gas+rad+rec EoS case.
- Update `test_sedov` so that test with radiation is compatible with current subroutines in `radiation_utils`.

Related issues:
- Resolves #654 (although radiative diffusion is not recommended for disks to begin with?)
- Relevant for #487


src/tests/test_eos.f90
```diff
@@ -326,10 +323,10 @@ subroutine get_rhoT_grid(npts,rhogrid,Tgrid)
 
  ! Initialise grids in Q and T (cgs units)
  npts = 30
- logQmin = -10.
```
themikelau, line 329: Check how much I can increase the Q and T range.

themikelau:
Pull request also fixes #654

src/tests/test_sedov.f90
```diff
@@ -108,12 +108,17 @@ subroutine test_sedov(ntests,npass)
     gmw      = 2.0
     if (do_radiation) then
        ! find for which T the function Etot*rho=Erad(T) + ugas(T)*rho is satified
-       temp = T_from_Etot(denszero,enblast,gamma,gmw)
+       denszero_cgs = denszero*unit_density
+       enblast_cgs = enblast*unit_ergg
+       temp = 0.
+       call get_idealplusrad_temp(denszero_cgs,enblast_cgs,gmw,temp,ierr)
```
danieljprice, line 114: minor thing but I'd prefer to keep a one-line function here instead of 4 lines that are hard to interpret

src/utils/analysis_common_envelope.f90
```diff
@@ -1505,57 +1503,68 @@ subroutine output_extra_quantities(time,dumpfile,npart,particlemass,xyzh,vxyzu)
...
+       endif
+    enddo
+enddo
 
- !Write data to file
- do i=1,1000
-    write(1002,"(1000(3x,es18.11e2,1x))") gam1_array(i,:)
- enddo
+ open(newunit=iunit,file='mesa_eos_pressure.out',status='replace')
+ write(iunit,"(1000(3x,es18.11e2,1x))") (pres_array(i,:), i=1,N)
```
danieljprice, line 1554: no need for the 1000() here, the pattern just repeats if you overflow it, so can just use (3x,es18.11e2,1x). Also I have never seen the e2 thing but I assume you know what it does

src/setup/set_star_utils.f90
```diff
@@ -513,38 +511,74 @@ subroutine set_star_thermalenergy(ieos,den,pres,r,npts,npart,xyzh,vxyzu,rad,eos_
...
-          vxyzu(4,i) = ugas_from_Tgas(tempi,gamma,gmw)
-          rad(iradxi,i) = radxi_from_Trad(densi,tempi)
-       else
-          vxyzu(4,i) = eni / unit_ergg
-       endif
+       tempi = min((3.*p_cgs/radconst)**0.25, p_cgs/(rho_cgs*Rg))  ! temperature guess
+       
+       select case(ieos)
+       case(2)
+          eos_type = 12  ! Calculate temperature from both gas and radiation pressure
```
danieljprice, line 521: I'd prefer to use a small function here as this kind of thing is very likely to be needed elsewhere. This is just temp_from_Etot, while the below is temp_from_Egas. If it can be cleaned up into a single line with a wrapper function that would be more maintainable (and less likely to be copy-pasted elsewhere)

src/utils/analysis_common_envelope.f90
```diff
@@ -1505,57 +1503,68 @@ subroutine output_extra_quantities(time,dumpfile,npart,particlemass,xyzh,vxyzu)
...
- integer :: i, j, ierr
- real    :: rho_array(1000) = (/ (10**(i/10000.), i=-180000,-30150,150) /)
- real    :: eni_array(1000) = (/ (10**(i/10000.), i=120000,149970,30) /)
- real    :: temp_array(400) = (/ (10**(i/1000.), i=3000,6990,10) /)
- real    :: kappa_array(1000,400)
- real    :: gam1_array(1000,1000)
- real    :: pres_array(1000,1000)
- real    :: dum(1000,1000)
- real    :: kappat, kappar
-
```
danieljprice, line 1520: this is a nice cleanup

danieljprice:
please note I have shifted where some of these options are read/written in the pending P-R on infile-cleanup; so if you want to merge this one first I will then resolve the conflicts

themikelau:
Ok, I will merge now and file issues for each of your suggestions.

src/main/eos_gasradrec.f90
```diff
@@ -20,91 +20,206 @@ module eos_gasradrec
...
- p = ( Rg*imu*d + radconst*T**3/3. )*T
- gamma_eff = 1.+p/(eint-d*erec)
- cf = sqrt(gamma_eff*p/d)
+ if (do_radiation_local) then
+    p = Rg*imu*d*T
+ else
+    p = ( Rg*imu*d + radconst*T**3/3. )*T
+ endif
+ if (present(xi)) then
+    cs2 = get_cs2(d,T,X,Y,do_radiation_local,xi)  ! not used at the moment
```
ryosuke-hirai, line 97: Why is this not used? Isn't this used if you do radiation transport?

PR #679, opened 2025-06-13
APR with GR

Fitz-Hu:
Type of PR: 
Bug fix / modification to existing code

Description:
Split particles along geodesics when using gr
Fix bugs with use apr with gr

Testing:
APR of TDE

Did you run the bots? yes


src/main/apr.f90
```diff
@@ -546,4 +618,119 @@ subroutine put_in_smallest_bin(i)
 
 end subroutine put_in_smallest_bin
 
+!-----------------------------------------------------------------------
```
danieljprice, line 619: as previously discussed, it seems very specific to push a particle forward on a geodesic trajectory by integrating the ODEs rather than just using the local force measurement to push the particle forwards in time. This routine looks ok in that it doesn't copy-and-paste large blocks of code from elsewhere, but are you sure it is necessary compared to just using the local acceleration estimate to extrapolate the velocity?

That is, how far forwards in "time" or "dist" does the integrate_geodesic_gr routine actually push the particle? Is this much more than would occur in one timestep or just a short extrapolation?

src/main/apr.f90
```diff
@@ -546,4 +618,119 @@ subroutine put_in_smallest_bin(i)
...
+       call externalforce(iexternalforce,xyz(1),xyz(2),xyz(3),xyzh(4), &
+                                timei,fext(1),fext(2),fext(3),poti,dt)
+       call externalforce_vdependent(iexternalforce,xyz,vxyz,fextv,poti,dens,uui)
+       fext = fext + fextv
+    endif
+ enddo
+
+ xyzh(1:3) = xyz(1:3)
+ vxyzu(1:3) = vxyz(1:3)
+end subroutine integrate_geodesic
```
danieljprice, line 732: one small issue with both of these routines is that they are not tested, if they are really needed [see above] then there should be unit tests for them... e.g. they could be compared to the output from calling substep_gr

Issue #683, opened 2025-06-25
isetdust=1 in DustyDisc setup not working anymore


RafaMartinezB:
While using SETUP=dustydisc, whenever I use in the .setup file:
            _isetdust =           1    ! how to set dust density profile (0=equal to gas,1=custom,2=equal to gas with cutoffs)_
So I can have a different dust distribution, no matter what changes I make to the file, if I run
_./phantomsetup mymodel.setup_ 
It will give me this message:
 _reading setup options from mymodel.setup
 opening database from mymodel.setup with 47 entries
  1 error(s) during read of setup file: re-writing...

 writing setup options file mymodel.setup_

I have tried everything, but I haven't been able to find what the code requires to make it work.

eragusa:
My bad....I had placed a read_inopt for initialising the dust density profile from file, but never really implemented it. That read_inopt looked for a line in the .setup file that was not there increasing nerr and causing the setup to fail. Rewriting did not write the missing line and phantomsetup continued not to work.
Removing the read_inopt for the missing line appears to have fixed the problem. 
I opened a pull request with the fix.

PR #686, opened 2025-06-29
(setup_disc) Fixed use_dustfile issue for isetdust=1,2

eragusa:
Description:
Fixed a line that caused an issue when initialising a dust disc with isetdust=1,2: I had placed a read_inopt for initialising the dust density profile from file. That read_inopt looked for a line in the .setup file that was not there increasing  the variable nerr and causing the setup to fail. Rewriting did not write the missing line and phantomsetup continued not to work.
Removing the read_inopt for the missing line appears to have fixed the problem.

Components modified:
Removed the faulty line

Type of change:
Bug fix


danieljprice:
Fixes #683

PR #693, opened 2025-07-08
Fix analysis common envelope star stability radius definitions

chunliangmu:
Description:
There seems to be several bugs in the radius calculations in the *star stabilization suite* in `phantomanalysis`.
For example, the `ipartrad` actually calculates the particle density radius instead of particle radius,
and the `ipdensrad`, instead of calculating the particle density radius, doesn't seem to calculate anything physical (because `iorder_a` is not actually sorting 'particles within surface by radius', since `xyzh` is not a sorted array; it is just sorting the first `npart_a` number of particles, which are not necessarily all within the surface);
also there seems to be a missing factor of 2 for the particle 2h radius calculations.
This results in some very confusing outputs, such as the calculated particle radius is smaller than the particle density radius, which shouldn't have happened because particle density radius only accounts for particles within the density surfaces, while particle radius accounts for all particles.
I also added a few lines to remove dead particles before doing the calculation.

Components modified:

- [ ] Setup (src/setup)
- [ ] Main code (src/main)
- [ ] Documentation (docs/)

Type of change:

- [x] Bug fix
- [ ] Physics improvements
- [ ] Documentation update
- [ ] Other (please describe)

Testing:

`make test` passed;
`make analysis` can compile;
results no longer looks erratic in terms of radius.

Did you run the bots? no

A test could check the analyzed solar profile (or other pre-computed star dump file that we know the actual radii of) etc outputs vs the correct answers.

Related issues: #

### DIFF
--- a/src/utils/analysis_common_envelope.f90
+++ b/src/utils/analysis_common_envelope.f90
@@ -1153,7 +1153,6 @@ subroutine star_stabilisation_suite(time,npart_in,particlemass,xyzh,vxyzu)
  npart = npart_in    ! npart might shrink in the process
  do i = 1,npart_in
     if (isdead(i)) then
-      xyzh(4,i) = 1.    ! Mark as alive to force re-killing and addition to dead list
       call kill_particle(i)
     endif
  enddo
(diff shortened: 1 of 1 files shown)


danieljprice:
just a comment that the factor is not always 2, as this depends on which kernel is being used. A safer way is to use the "radkern" parameter from the kernel module, which is what is done in the main code

chunliangmu:
> just a comment that the factor is not always 2, as this depends on which kernel is being used. A safer way is to use the "radkern" parameter from the kernel module, which is what is done in the main code

Thanks Daniel! I just want to clarify that I hardcoded the factor of 2 because the names of the output variables explicitly have `2h` in their names (see original file line 1160 'part 2h rad' and 1162 'p2h dens rad')

PR #711, opened 2025-09-03
Remove every refs to the deprecated linked list

Yrisch:
Description:

This PR removes or renames every variable, comment, file name, module name, routine name or doc that references the deprecated linked list used to find neighbours. This linked list is not used anymore and has been replaced by the actual cached, tree-based neighbour finding. It will clarify the code reading in the future and avoid confusion for further development on the tree. 

I'm planning to add a secondary tree using only sink particles to optimise the accretion loop when the sink number starts to go higher than a thousand. It was then necessary to do this cleaning before doing any modifications to the tree routines.

The major changes:
- `linklist_kdtree.f90` -> `neigh_kdtree.f90` file name
- `linklist`                     -> `neighkdtree`         module name
- `set_linklist`              -> `build_tree`             routine name
- `ifirstincell`               -> `icell_is_active`                variable name
- `test_link.F90`          -> `test_neigh.f90`     file name
- `testlink`                   -> `testneigh`              module name
- `itimer_link`             -> `itimer_tree`            variable name

Components modified:

- [x] Setup (src/setup)
- [x] Main code (src/main)
- [x] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [ ] Physics improvements
- [x] Documentation update
- [x] Other : rm deprecated linkedlist refs

Testing:
It is just a big rename operation; nothing should be broken. Tests should all be green if I missed nothing.

Did you run the bots? yes

### DIFF
--- a/src/main/apr.f90
+++ b/src/main/apr.f90
@@ -429,7 +429,7 @@ end subroutine splitpart
 !-----------------------------------------------------------------------
 subroutine merge_with_special_tree(nmerge,mergelist,xyzh_merge,vxyzu_merge,current_apr,&
                                      xyzh,vxyzu,apr_level,nkilled,nrelax,relaxlist,npartnew)
- use neighkdtree,   only:build_tree,ncells,ifirstincell,get_cell_location
+ use neighkdtree,   only:build_tree,ncells,itypecell,get_cell_location
  use mpiforce,      only:cellforce
  use kdtree,        only:inodeparts,inoderange
  use part,          only:kill_particle,npartoftype,igas
@@ -456,7 +456,7 @@ subroutine merge_with_special_tree(nmerge,mergelist,xyzh_merge,vxyzu_merge,curre
  ! be merged or not
  com = 0.
  over_cells: do icell=1,int(ncells)
-    i = ifirstincell(icell)
+    i = itypecell(icell)
     if (i == 0) cycle over_cells !--skip empty cells
     n_cell = inoderange(2,icell)-inoderange(1,icell)+1
 
--- a/src/main/apr.f90
+++ b/src/main/apr.f90
@@ -15,7 +15,7 @@ module apr
 ! :Runtime parameters: None
 !
 ! :Dependencies: apr_region, dim, get_apr_level, io, io_summary, kdtree,
-!   neighkdtree, mpiforce, part, physcon, quitdump, random, relaxem,
+!   mpiforce, neighkdtree, part, physcon, quitdump, random, relaxem,
 !   timestep_ind, utils_apr, vectorutils
 !
  use dim, only:use_apr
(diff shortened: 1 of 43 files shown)


src/main/apr.f90
```diff
@@ -450,13 +450,13 @@ subroutine merge_with_special_tree(nmerge,mergelist,xyzh_merge,vxyzu_merge,curre
...
 
- call set_linklist(nmerge,nmerge,xyzh_merge(:,1:nmerge),vxyzu_merge(:,1:nmerge),&
+ call build_tree(nmerge,nmerge,xyzh_merge(:,1:nmerge),vxyzu_merge(:,1:nmerge),&
                       for_apr=.true.)
  ! Now use the centre of mass of each cell to check whether it should
  ! be merged or not
  com = 0.
  over_cells: do icell=1,int(ncells)
-    i = ifirstincell(icell)
+    i = itypecell(icell)
```
danieljprice: love the enthusiasm, however the renaming of this variable is not correct, it really is the "first in cell" index, and we use the ll(i) to find the next particle in the cell. The index does not determine the type of the cell in any way, but the index can be negative to indicate that the cell is inactive

src/main/dens.F90
```diff
@@ -1161,7 +1161,7 @@ subroutine reset_neighbour_stats(nneightry,nneighact,maxneightry,maxneighact,nca
  maxneighact = 0
  ncalc = 0_8
  nneighact = 0
- nrelink = 0_8
+ nrengh = 0_8
```
danieljprice, line 1164: can we have a more intelligible variable name? like n_redo_neighbours ? or ncalls?

src/main/kdtree.F90
```diff
@@ -1589,7 +1588,7 @@ subroutine revtree(node, xyzh, ifirstincell, ncells)
 
 !$omp parallel default(none) &
 !$omp shared(maxp,maxphase) &
-!$omp shared(xyzh, ifirstincell, ncells, apr_level) &
+!$omp shared(xyzh, itypecell, ncells, apr_level) &
 !$omp shared(node, ll, iphase, massoftype, maxlevel,aprmassoftype) &
 !$omp private(hmax, r2max, xi, yi, zi, hi, il, ir, nodel, noder) &
 !$omp private(dx, dy, dz, dr2, icell, i, x0) &
```
danieljprice, line 1594: all ok, but please revert naming of ifirstincell

danieljprice (commented):
looks good, just please revert ifirstincell renaming

src/main/dens.F90
Yrisch: Absolutely, Gemini told me the same.

src/main/apr.f90
Yrisch: I thought it was not the case anymore? In `force` and `dens` we use inoderange to cycle over the cell's particles. But I may have missed something

danieljprice: I think you are right :) In this case we should get probably rid of ifirstincell altogether and use inodepart, we just need a way to tag if the whole cell is active or inactive

Yrisch: Cool, let's do it that way then!

Yrisch: Regarding this variable, there are some obsolete routines that we may want to update or remove, such as `revtree` or `sort_part`. They are never used in the current code base, but it may be dangerous to leave them broken in the code.

PR #723, opened 2025-09-19
Apr file tracking and code update

becnealon:
Description:
Added APR region tracking files. File prints the APR region location in x, y, z as well as the radii of the different refinement levels. File also includes the time stamp and the dumpfile number that it corresponds to.
Also tidied up some of the new options and removed the ran2 in the particle splitting code.

Components modified:
- [x] Main code (src/main)

Type of change:
- [x] Performance improvements
- [x] Code cleanup / refactor
- [x] New feature

Testing:
make DEBUG=yes

Did you run the bots? no

Did you update relevant documentation in the docs directory? N/A

Is there a unit test that could be added for this feature/bug? N/A

### DIFF
--- a/src/main/utils_apr.f90
+++ b/src/main/utils_apr.f90
@@ -292,4 +292,85 @@ subroutine write_options_apr(iunit)
 
 end subroutine write_options_apr
 
+!-----------------------------------------------------------------------
+!+
+!  Writes tracking file for APR regions
+!+
+!-----------------------------------------------------------------------
+
+subroutine write_aprtrack(tdump,dumpfile)
+ use io, only:iaprdump,error
+ real,             intent(in) :: tdump
+ character(len=*), intent(in) :: dumpfile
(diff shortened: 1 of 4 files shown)


src/main/utils_apr.f90
```diff
@@ -292,4 +292,79 @@ subroutine write_options_apr(iunit)
...
+ logical :: iexist
+
+
+ if (ntrack == 0) return ! nothing to do here
+
+ ! clever formatting
+ dump_length = len_trim(dumpfile)
+ start_pos = dump_length - 5 + 1
+ read(dumpfile(start_pos:dump_length), *) dumpfile_int
+ ! dynamically make the formatting string to accomodate the right number of regions
```
danieljprice, line 320: house style is to NOT SHOUT in format strings, i.e. lowercase

danieljprice (approved):

Issue #712, opened 2025-09-04
Sink in tree should be a runtime option


Yrisch:
There is no need for a compile-time option for this. to be fixed

Yrisch:
Done in #727

PR #727, opened 2025-09-23
Sink in tree runtime parameter 2

Yrisch:
Description:
revision of #726  without the miss commit of a dumpfile: This PR remove the precompile flag SINKTREE that wasn't necessary—related to #55. I had to perform a turnaround to correctly reallocate memory during tests, as sinktree needs to adjust the size of the main arrays.

Components modified:

- [x] Main code (src/main)
- [x] Test suite (src/tests)

Type of change:

- [x] Code cleanup / refactor

Testing:
No change in tests expected.

Did you run the bots? no

Did you add comments such that the purpose of the code is understandable? no

Is there a unit test that could be added for this feature/bug? no

Related issues: #55 #712


src/tests/testsuite.f90
```diff
@@ -454,6 +461,21 @@ subroutine testsuite(string,first,last,ntests,npass,nfail)
...
 !
+!--test of sink in tree
+!
+ use_sinktree = dosinktree
+ if ((dosinktree.or.testall) .and. use_sinktree) then
+    if (use_sinktree) then
+       maxp_old = maxp
+       call deallocate_memory()
+       call allocate_memory(int(maxp_old,kind=8))
+    endif
```
danieljprice: these 3 lines could be put into a new routine called reallocate_memory(). Right now it would just do a deallocate and allocate with the same array size (but with a config option changed), but in principle it could be more clever and only allocate arrays that are not already allocated

danieljprice: another option would be to make use_sinktree = .true. when allocating memory in the test suite, then set it to false for the first set of tests, then set it back to true later. That way the memory would be allocated in the first place but the tests would run with use_sinktree = .false.

danieljprice (approved):

src/tests/testsuite.f90
Yrisch: Agree with both options. The second one could work. But the first allocation is done in the `phantomtest`, where there are no option checks. It would be a bit ugly to put an if statement here. I'll try the first one!

Yrisch: I added a reallocation option in `allocate_memory` to only have one line in the test if statement. Should be good to merge if tests pass

PR #734, opened 2025-10-02
Infile parsing uses read_inopt routines, like reading .setup file

### LINKED ISSUE #594: read .in file using the routines used to read .setup files
For #513 it would be helpful to read the phantom infile using the database/read_inopts routines that are used to read the .setup file

This should have been done anyway but was always a bit too "big and complicated" to do because there are so many modules using the old style read_options routines with select case and read statements... but it's time to clean this up

_Originally posted by @danieljprice in https://github.com/danieljprice/phantom/pull/513#discussion_r1787207652_

danieljprice:
Description:
A big job on the backlog was to address #594, namely to use the nice api routines to read and parse options from the .in file the same way as the .setup file is handled. This was always the intended use of these routines, but just never got around to doing this.

- cleaned up all read_options routines to use the new format
- made minor changes to the options parsing in some modules (e.g. logical instead of integer for 0/1 flag)
- adding missing comment headers to subroutines
- read_options routines now better reflect write_options routines
- avoid duplicated code in libphantom_amuse, now calls the same routines as in the main code with no duplication

Components modified:

- [ ] Setup (src/setup)
- [x] Main code (src/main)
- [ ] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [ ] Physics improvements
- [ ] Documentation update
- [ ] Better testing
- [x] Code cleanup / refactor
- [ ] Other (please describe)

Testing:
Parsing of the .in file seems to work as previously, including loop syntax and warning about unknown/non-retrieved variables present in the .in file

Did you run the bots? yes

- should be better testing of option combinations for the .in and .setup files. This is partially checked by running the test suite
- some of the swap-in modular routines are not compiled anywhere by default, so the build for these should be checked (e.g. SETUP=test could try to compile all the inject_* modules, similarly for moddump_*, analysis_* etc; dependencies of these modules may be an issue however)

Related issues: Closes #594


src/main/radiation_utils.f90
```diff
@@ -121,41 +121,24 @@ end subroutine write_options_radiation
...
- igotall = (ngot >= 1)
+subroutine read_options_radiation(db,nerr)
+ use dim,          only:store_dust_temperature
+ use eos,          only:iopacity_type
+ use infile_utils, only:inopts,read_inopt
+ type(inopts), intent(inout) :: db(:)
+ integer,      intent(inout) :: nerr
+
+ call read_inopt(implicit_radiation,'implicit_radiation',db,errcount=nerr)
+ if (implicit_radiation) store_dust_temperature = .true.
```
themikelau, line 132: @danieljprice  Is there any particular reason to store dust temperature when using implicit_radiation, even if there is no dust in the simulation?

PR #737, opened 2025-10-06
unify orbital dynamics helper routines into single module

### LINKED ISSUE #736: consolidate orbital dynamics helper functions into a single module
Currently there are at least 4 modules that all have essentially the same function:

- utils_kepler.f90 : utility routines for Keplerian orbits
- utils_orbits.f90: utility routines for orbital dynamics
- utils_binary.f90 : utility routines for setting up binaries
- set_binary.f90 : also contains a bunch of utility functions

these should be consolidated into a single set of tested and documented routines in ONE module...

danieljprice:
Description:
Addresses #736 by consolidating all orbital dynamics routines into a single module (utils_orbits.f90)
- moved utils_orbits.f90 -> src/main instead of src/utils
- added test_orbits module containing unit tests of these routines
- tried to use common logic and variable names inside all routines where possible. Some differences remain.
- preliminary work on a "flyby reconstructor" set of orbital parameters, supplying observed separation & velocity at t=tobs, and a desired initial separation from which to start the simulation. This is unit tested but needs to be implemented as an option in setup_disc.f90
- added a few more helper functions (e.g. get_inclination)
- used generic interfaces for routines giving the same output but with different input variables (e.g. eccentricity from dx,dv)
- simplified units handling by using the standard gravitational parameter mu = G*(m1+m2) in all routines

Components modified:

- [x] Setup (src/setup)
- [x] Main code (src/main)
- [x] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [x] Physics improvements
- [x] Better initial conditions
- [ ] Documentation update
- [x] Better testing
- [x] Code cleanup / refactor
- [ ] Other (please describe)

Testing:
New unit tests in test_orbits.f90 module now part of the test suite

Did you run the bots? no

already implemented

Related issues: fixes #736


Yrisch (approved):
Seems all good to me. Thanks for the cleaning!

PR #740, opened 2025-10-08
Enhancement and bug fix related to radiation

themikelau:
- Write maximum errors in radiation and gas energies into ev file.
- Bug fix: The timing variables in each implicit radiation iteration, `t1` and `tcpu1` , need to be declared as shared and not private variables.


src/main/energies.f90
```diff
@@ -627,6 +629,13 @@ subroutine compute_energies(t)
        endif
     enddo
     !$omp enddo
+
+    !$omp single
+    if (do_radiation .and. implicit_radiation) then
+       call ev_data_update(ev_data_thread,iev_errE,rad_errorE)
+       call ev_data_update(ev_data_thread,iev_errU,rad_errorU)
+    endif
+    !$omp end single
```
themikelau, line 638: @danieljprice Is this the correct way and place to do this?

danieljprice: Looks ok, depends if you need a reduction over mpi threads or not. Right now I think it's ok as implicit radiation does not currently work with MPI anyway...

danieljprice (approved):

PR #765, opened 2025-11-10
APR in MCFOST

JotaRata:
Description:
Pass `apr_level` as an optional argument to `run_mcfost_phantom`.
~Requires https://github.com/cpinte/mcfost/pull/148~
Requires https://github.com/cpinte/mcfost/pull/151

Components modified:

- [ ] Setup (src/setup)
- [ ] Main code (src/main)
- [ ] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [x] Physics improvements
- [ ] Documentation update
- [ ] Better testing
- [ ] Code cleanup / refactor
- [x] Other (please describe)

Testing:

> Ran test Phantom simulations with MCFOST=yes APR=yes.
> Temperature profile matches the control even inside the APR regions.

Did you run the bots? yes

Did you add comments such that the purpose of the code is understandable? no

Is there a unit test that could be added for this feature/bug? no

~There are still bugs to be fixed, per example, Phantom freezes after deleting dump_00000.tmp~

### DIFF
--- a/src/utils/analysis_mcfost.f90
+++ b/src/utils/analysis_mcfost.f90
@@ -69,6 +69,7 @@ subroutine do_analysis(dumpfile,num,xyzh,vxyzu,particlemass,npart,time,iunit)
  logical, parameter :: write_T_files = .false. ! ask mcfost to write fits files with temperature structure
  character(len=len(dumpfile) + 20) :: mcfost_para_filename
  real :: a_code,rhoi,pmassi,Tmin,Tmax,default_kappa,kappa_diffusion
+ integer(kind=1), dimension(npart) :: new_level
 
  if (.not. use_mcfost) return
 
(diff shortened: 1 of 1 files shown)


JotaRata:
Rebase notes: I rewrote the commits to not sabotage the git blame history and to incorporate the changes from https://github.com/cpinte/mcfost/pull/151 which is the current merge candidate for the APR implementation in MCFOST.

I also removed the conditional logic around calling MCFOST’s init/run routines based on whether Phantom was compiled with APR. This breaks backwards compatibility and it *will* crash if the caller and callee interfaces mismatch.

also why does gemini hates me:(

danieljprice:
now that https://github.com/cpinte/mcfost/pull/151 is merged into MCFOST I am happy to merge this, however there is still a compilation failure on the Phantom side that needs to be fixed (related to the use_apr flag)

danieljprice:
mcfost took more than 6 hours to complete (should normally be 5 mins) so the job timed out. This smells like a bug to me...

JotaRata:
Also, according to the logs, the first run was completed in one minute and after it's completed it calls mcfost again and it hangs, which is weird..

I reported this behavior a while ago and it was fixed, now it seems it's back.

JotaRata:
I downloaded and used the exact parameters as in the test and was unable to reproduce this on my machine.

I did have a "Floating-point exception" when compiling phantom with DEBUG=yes and the backtrace pointed at  line 129:
https://github.com/JotaRata/phantom/blob/bfb0cfad5e61606ef4d582bdaad5b9943645f1b1/src/utils/analysis_mcfost.f90#L129

Could be that MCFOST expects `apr_level` to have a size equal to `npart` as defined here (line 192):
https://github.com/cpinte/mcfost/blob/9805b5a2c858fe453c0e4e851a0d816a5d23cc39/src/mcfost2phantom.f90#L192

danieljprice:
> Two tests fail with a message "fatal: No names found, cannot describe anything." that appears to be generated by git

The last failures were just because of download failures so rerunning those tests fixes them

PR #807, opened 2026-03-20
(inject_wind) fix bug with particle mass estimate

lsiess:
Description:
An error was introduced in the expression of the wind resolution. 

Components modified:

- [ ] Setup (src/setup)
- [x] Main code (src/main)
- [ ] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [x] Bug fix
- [ ] Physics improvements
- [ ] Documentation update
- [ ] Better testing
- [ ] Code cleanup / refactor
- [ ] Other (please describe)

Testing:

Did you run the bots?  no

Did you update relevant documentation in the docs directory?  no

Did you add comments such that the purpose of the code is understandable?  no

Is there a unit test that could be added for this feature/bug?  no

Related issues: #


danieljprice:
@lsiess we found the issue with the test suite, if you pull off main once #809 is merged it will be fixed

lsiess:
There is a problem with `test_wind`. When dust acceleration is considered, the Bernoulli energy cannot remain constant as defined because the gravitational energy term must include the reduction of the effective gravitational potential caused by the dust acceleration (the Eddington factor). The problem is that this factor depends on the position so we cannot have an accurate value for this potential. 
Also, it seems that the boundary particles got reassigned a gas type ? This may be correct but it makes the .ev file unreadable because it is expected to output a boundary particle property that is not defined anymore.

danieljprice:
@lsiess please can you mark this as "ready for review"? Also do you want me to resolve the conflicts with main or happy to do this from your end?

PR #815, opened 2026-05-18
Get setting up a star with variable composition working again

themikelau:
- Add conflict error when using MESA EoS with variable composition (since tables have fixed composition). 
- Adjusted logic for determining whether to write X, Z, mu and gamma to full dumps.
- Read mean molecular weight from MESA profile for setting up star with variable composition.
- Fixed issue with reading MESA profiles caused by incorrectly determining the number of header lines.
- Set `eos_vars(imu,:)` in `set_star_composition`.


danieljprice:
@themikelau any chance we can add a unit test for this to src/tests/test_setstar.f90?

themikelau:
We absolutely should. I am still working on this pull request, including a fix to `read_mesa`, which no longer seems to read the correct number of header lines (6) in a MESA profile.

src/main/readwrite_dumps.f90
```diff
@@ -228,20 +228,19 @@ subroutine write_fulldump(t,dumpfile,ntotal,iorder,sphNG)
        if (eos_outputs_temp(ieos) .or. (.not.store_dust_temperature .and. icooling > 0)) then
           call write_array(1,eos_vars,eos_vars_label,1,npart,k,ipass,idump,nums,nerr,index=itemp)
        endif
-       if (eos_is_non_ideal(ieos)) call write_array(1,eos_vars(igamma,:),eos_vars_label(igamma),npart,k,ipass,idump,nums,nerr)
```
themikelau, line 231: @danieljprice What condition do we really want for the writing of gamma? This is clearly not correct as non-ideal EoSs are often used with fixed gamma (e.g. MESA, ideal gas + radiation).

src/setup/set_star_utils.f90
```diff
@@ -447,15 +445,14 @@ subroutine set_star_composition(use_var_comp,use_mu,npart,xyzh,Xfrac,Yfrac,&
...
+ !$omp shared(Xfrac,Yfrac,mtab,mu,eos_vars) &
  !$omp private(i,massri)
  do i = i1+1,npart
-    massri = mass_enclosed_r(i-i1)/Mstar
-    if (use_var_comp) then
-       eos_vars(iX,i) = yinterp(Xfrac,mtab,massri)
-       eos_vars(iZ,i) = 1. - eos_vars(iX,i) - yinterp(Yfrac,mtab,massri)
-    endif
+    massri = mass_enclosed_r(i-i1)
+    eos_vars(iX,i) = yinterp(Xfrac,mtab,massri)
```
themikelau, line 453: Removed check for use_var_comp since this subroutine is only called when use_var_comp=.true.

danieljprice:
Yikes, we should add a unit test for that too. The setup tests do check that all combinations of the iprofile option work when running phantomsetup with SETUP=star, so I'm surprised this broke. Does it need testing on a different file ?

src/main/readwrite_dumps.f90
danieljprice: A logical routine called Eos_outputs_gamma would be consistent with the logic around eos_outputs_temp

themikelau:
@danieljprice I will add a unit test in a separate pull request.
@Yrisch Could you possibly check that the changes I have made to `readwrite_dumps` are still compatible with usage with krome? Thanks.

Yrisch:
I'm never using Krome in my setups. That's a mistake in PR #814. I should have cleaned that. You can remove the (ieos/=22) condition.

themikelau:
@danieljprice The reason why the MESA profile reading bug has been undetected is because the default stellar profile is `P12_Phantom_Profile.data`, which is not a MESA profile... I will add a test to reproduce the Lau+22a,b star.

danieljprice:
I was thinking the default should really be a solar type star, have asked Fitz to upload the one used in Hu et al TDEs papers, which would be a more sensible default

themikelau:
Sure. And the default for MESA profiles should really be a MESA profile. We should still include a separate test for reading `P12_Phantom_Profile.data` since this is the file format used for softened stellar profiles.

PR #817, opened 2026-05-25
Add test for star setup with radiation

themikelau:
Checks star can be set up with radiation and that Tgas = Trad initially. Skips EoSs that are incompatible with radiation.


danieljprice:
All looks good to me

PR #822, opened 2026-06-03
Added element input to eos helmholtz with default being C/O 50,50. 

AliPourmand:
Description:
The element abundance in EOS helmholtz was hardcoded before; now elements H,He,C,O,Ne,Mg can be added as a runtime parameter in .in files when eos is helmholtz

Components modified:

- [ ] Setup (src/setup)
- [ x] Main code (src/main)
- [ ] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [x ] Bug fix
- [ ] Physics improvements
- [ ] Documentation update
- [ ] Better testing
- [x ] Code cleanup / refactor
- [ ] Other (please describe)

Testing:
 setting the total elements to more than 1 and seeing if it gives errors

Did you run the bots? no

Did you add comments such that the purpose of the code is understandable? yes/

### DIFF
--- a/src/main/eos.f90
+++ b/src/main/eos.f90
@@ -1724,7 +1724,7 @@ end subroutine read_headeropts_eos
 subroutine write_options_eos(iunit)
  use dim,            only:use_krome,isothermal,mhd
  use infile_utils,   only:write_inopt
- use eos_helmholtz,  only:eos_helmholtz_write_inopt
+ use eos_helmholtz,  only:write_options_eos_helmholtz
  use eos_barotropic, only:write_options_eos_barotropic
  use eos_piecewise,  only:write_options_eos_piecewise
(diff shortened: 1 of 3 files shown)


danieljprice:
the test failure is the same one we had on Antoine's pull request, seems a bit stochastic but likely related to previous changes. I suspect the temp(:) array we added needs to be at least allocated/initialised to zero for the non-Helmholtz cases

AliPourmand:
I added a line in `read_star_profile` after allocation to set temp=0 for all non-file based profiles, would that be a good fix?

```
 allocate(r(ng_max),den(ng_max),pres(ng_max),temp(ng_max),en(ng_max),mtab(ng_max))
 temp = 0.  ! Initialize temperature array for non-file-based profiles
```

PR #858, opened 2026-07-29
Aton reader module added (similar to MESA reader)

orsolademarco-astronomy:
Description:
Added a readwrite_aton module similar to MESA and KEPLER. Updated all the lines in setup that needed to include the new option.

Components modified:
- [ X] Setup (src/setup)
- [ ] Main code (src/main)
- [ ] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [ ] Physics improvements
- [ ] Documentation update
- [ ] Better testing
- [ ] Code cleanup / refactor
- [X ] Other Added a new module like readwirte_mesa.f90 called readwrite_aton.f90 to read star files from another code.

Testing:
I can read the file which is now option 8, and the behaviour is identical to mesa.

Did you run the bots? no

Is there a unit test that could be added for this feature/bug? yes I think anything that is done for MESA would work.

a test should check that running phantomsetup on SETUP=star with ieos=10 and option=8 successfully creates a dump

## Summary by CodeRabbit

* **New Features**
  * Added support for importing/exporting ATON-format stellar profiles.
  * Introduced ATON as a selectable density-profile/EOS option, including input-profile handling and updated setup/option I/O.

* **Bug Fixes**
  * Improved ATON profile detection and validation by inferring ATON from column labels and expanding nonnegativity checks to cover additional required fields.
  * Refined softened-core workflows to reuse ATON read/write steps consistently.
  * Enhanced multi-star EOS setup so variable-component options are requested only when relevant.

* **Chores**
  * Updated build configuration to include an additional ATON-related compilation unit.
  * Updated editor settings to disable the Fortran language server.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -634,7 +634,7 @@ libsetup: $(OBJLIBSETUP)
 .PHONY: phantomsetup
 phantomsetup: setup
 
-SRCSETSTAR= prompting.f90 density_profiles.f90 readwrite_kepler.f90 readwrite_mesa.f90 \
+SRCSETSTAR= prompting.f90 density_profiles.f90 readwrite_kepler.f90 readwrite_mesa.f90 readwrite_aton.f90 \
             set_cubic_core.f90 set_fixedentropycore.f90 set_softened_core.f90 \
             set_star_utils.f90 relax_star.f90 set_star.f90 set_orbit.f90
 
--- a/src/.vscode/settings.json
+++ b/src/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "fortran.fortls.disabled": true
+}
--- a/src/setup/readwrite_aton.f90
+++ b/src/setup/readwrite_aton.f90
@@ -14,7 +14,7 @@ module readwrite_aton
 !
 ! :Runtime parameters: None
 !
-! :Dependencies: datafiles, fileutils, physcon, units
+! :Dependencies: datafiles, fileutils, physcon, units, table_utils
 !
  implicit none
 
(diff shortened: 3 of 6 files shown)


orsolademarco-astronomy:
Hello Daniel, I have addressed all the bot comments that came to me after my pull request. I have tested and re-committed and also pushed.

orsolademarco-astronomy:
One more round of minor fixes identified by the bot.

danieljprice:
Delighted to see your pull request @orsolademarco-astronomy :)

One thing that would be helpful for testing is to supply an example Aton file that we can use as the default setup for iprofile=8. You should be able to just attach it as a file in the description or comments thread, or alternatively post an example file on zenodo and link to the phantom zenodo community (and update src/main/datafiles.f90 with the lookup address). 

One of the tests we run is to check that default SETUP=star with iprofile=[0,1,2,3,4,5...] etc all manage to relax a star with default settings. This is done by some code in scripts/buildbot.sh:
```
test_setupfile_options()
{
   myfail=0;
   setupfile=$1;
   flags=$2;
   setup=$3;
   infile=$4;
   range=''
   if [ "X$setup" == "Xstar" ]; then
      param='iprofile1'
      range='0 1 2 3 4 5 6 7'
   fi
 ```

It'd make sense to expand the range to 8... that way we can guarantee that the ATON read is not broken. 

There are also a bunch of tests run in .github/workflows/binary.yml that test against various "known" .setup and .in files so could equivalently add something there (e.g. if you want to test something other than a single star)

orsolademarco-astronomy:
I have a question about updating datafiles.f90. I expected to have to add the new zenodo repo, but what I found was there is only one line referring to data/star_data_files which currently points to a zenodo repo that Ali has added with his WD file. (I realise that a number of other star data files are inside the phantom repo, like the Kpler or Jan stars.) If I just want to add my new zendo repo with the ATON star what case should I use since "my case" is already used? This is what is in the file:

case('data/star_data_files')
    url = 'https://zenodo.org/records/20738843/files/'

but I also need the file to go into data/star_data_files, but I cannot make a new case, nor have another URL in there. The AI suggestion is a structural rewrite of datafiles.f90 to be more granular and search files rather than entire directories, but I do not really want to go there. 

So the alternative you give is the zenodo link to phantom which I have done. But the other suggestions you make "You should be able to just attach it as a file in the description or comments thread" seems vague. Should I not just commit the file to the data_star_file for now? Awaiting a rationalisation of star files in general?

danieljprice:
I will merge this for now, I can add the test in a subsequent P-R, the file is online so should be easy.

General documentation on datafiles here:
https://phantomsph.readthedocs.io/en/latest/developer-guide/datafiles.html

PR #861, opened 2026-08-05
Boost factor for gravitational-wave inspiral

themikelau:
- Add boost factor to gravitational radiation reaction force, so that this external force option can be used more generally as orbit damping.
- Configure this external force option for `moddump_binary`.
- Deletes `moddump_sinkbinary` (#862).

https://github.com/user-attachments/assets/3015bcde-d822-4ea8-b0dc-39d465111b18

The attached video shows an example application, where the left panel shows the use of the gravitational wave inspiral force with a boost factor of ≈ 4*10^11 to achieve an orbital reduction of da/dt ≈ 1 Rsun / orbit. One could also interpret this as slowing the speed of light by a factor of ≈ 200, because da/dt scales as c^-5.

## Summary by CodeRabbit

* **New Features**
  * Added optional gravitational-wave inspiral support for binary setup and dump workflows.
  * Added a configurable multiplier for gravitational-wave forces.
  * Binary configurations now preserve inspiral settings and star particle counts.

* **Bug Fixes**
  * Improved star particle tracking and centre-of-mass calculations for inspiral systems.
  * Added validation that gravitational-wave forces use compatible code units.

* **Changes**
  * Removed the legacy sink-binary dump modification workflow.


danieljprice:
build failure on binary is because new options should not in general be compulsory in the .setup file (and thus require a rewrite of old .setup files)

can fix with read_inopt(..., default=gwinspiral) which will give a default value if not found in the .setup file

danieljprice:
for the nstar1 and nstar2 stuff, this is kind of horrid as it uses the particle ids to "remember" which star the particle belonged to in the first place so as to apply the radiation reaction from a two body problem to the entire collection of particles belonging to "star 1" and "star 2".

At worst we should at least use the iorig array here as a better track of the original particle identities (e.g. would work across MPI threads). In general it'd be nicer to think of how to apply gravitational wave emission in a less hokey way... for example we already have routines to compute the emission by computing the total time varying quadrupole moment , presumably wouldn't be too hard to apply a backreaction force based on that 

Final suggestion would be if setting gwinspiral to true then you should automatically turn on the gravitational wave strain calculation in the .in file (gravitationalwaves= T) so you get the "actual" gravitational wave emission calculated in the .ev file for comparison)

themikelau:
Thanks for the comments, Daniel. The reason why I looked into this is because when doing large mass-ratio CEs or stellar mergers, the transition from a pseudo-Keplerian orbit to a dynamical plunge-in could take a long time, wasting loads of CPU time. And so, following what the AREPO folks have been doing for CE and white-dwarf merger simulations (Pakmor+21, https://arxiv.org/abs/2103.06277), I have decided to apply a secular damping term to the orbit. The easiest way of doing this seems to be boosting the existing gravitational radiation reaction. Doing this by calculating the time-varying quadrupole moment would be an overkill for my purposes, although it may be useful for someone else.

PR #874, opened 2026-09-11
angular momentum conserving tree

danieljprice:
Implemented the correction term from Marcello (2017; https://ui.adsabs.harvard.edu/abs/2017AJ....154...92M) which gives exact angular momentum conservation in the tree

Description:
The new algorithm requires storing octupole moments during the tree build. On short test supplied by @s-neilson the performance hit is about 5% which seems acceptable.

Components modified:

- [ ] Setup (src/setup)
- [x] Main code (src/main)
- [x] Test suite (src/tests)
- [ ] Documentation (docs/)

Type of change:

- [ ] Bug fix
- [x] Physics improvements
- [ ] Documentation update
- [x] Better testing
- [ ] Code cleanup / refactor
- [ ] Other (please describe)

Testing:
Added new unit test for exact angular momentum conservation in the tree

Did you run the bots? no

Did you add comments such that the purpose of the code is understandable? yes/no

already added one

Related issues: #

## Summary by CodeRabbit

* **Enhancements**
  * Improved gravitational calculations by incorporating octupole moments alongside quadrupole data.
  * Added torque corrections to improve angular momentum conservation.
  * Improved parallel handling of higher-order gravitational data.
  * Ensured tree initialization and rebuilding correctly process octupole information.

* **Tests**
  * Expanded self-gravity validation to check linear and angular momentum conservation.
  * Updated tree-rebuild validation to cover octupole moments.

image_720


src/main/kdtree.F90
```diff
@@ -1814,6 +1824,85 @@ pure subroutine compute_M2L(dx,dy,dz,dr1,q0,quads,fnode)
 
 end subroutine compute_M2L
 
+#ifdef GRAVITY
```
Yrisch, line 1823: No ifdef pocily. This one can be avoided.

src/main/kdtree.F90
```diff
@@ -1711,14 +1720,15 @@ subroutine node_interaction(node_dst,node_src,tree_acc2,fnode,stackit,xoffset,yo
...
+       call compute_M2L(dx,dy,dz,dr1,node_src%mass,node_src%quads,fnode)
+       call add_torque_correction(dx,dy,dz,dr1,node_dst%mass,node_src%mass, &
+                                  node_dst%octs,node_src%octs,fnode)
 #else
-       mass_src=0.
-       quads_src=0.
+       quads_zero = 0.
+       call compute_M2L(dx,dy,dz,dr1,0.,quads_zero,fnode)
 #endif
-       call compute_M2L(dx,dy,dz,dr1,mass_src,quads_src,fnode)
```
Yrisch: ComputeM2L is not necessary if gravity is off. I suggest pushing the whole if(.not. cache) clause into the ifdef. Remove the fallback with quads_zero and just set stackit to false if gravity off. But if I'm correct use_dualtree is set to false if gravity is off. (It should be)

src/main/kdtree.F90
```diff
@@ -1692,7 +1698,10 @@ subroutine node_interaction(node_dst,node_src,tree_acc2,fnode,stackit,xoffset,yo
  logical,      intent(out)   :: stackit
  real    :: dx,dy,dz,r2,dr1
  real    :: rcut_dst,rcut_src,rcut,rcut2
- real    :: size_dst,size_src,mass_src,quads_src(6)
+ real    :: size_dst,size_src
+#ifndef GRAVITY
+ real    :: quads_zero(6)
+#endif
```
Yrisch: Could be avoided with next comment

Yrisch (changes requested):
That's very cool ! I have a few questions.
These correction terms are directly applied to the first three fnodes. It seems independent of the order of expansion. Could it be worth to try raise the M2L expansion to order 4 now that we pay for the octupole? (Maybe another PR)
How did you benchmark the perf hit? Is it only on the force or also on the tree build? I'm working on the ultimate caching strategy to prune the majority of the thread work duplication (tree walk + M2L). This might erase most of the perf hit you measure here.

danieljprice:
It's a good question whether we should use the octupoles now we have them, this is just implementing the correction direct from the Marcello paper so it's indeed fixed at P=3. I suspect the problem is that if we used the octupoles to get better accuracy we'd need another set of moments to get back AM conservation...

danieljprice:
For the performance hit I benchmarked on a 50k binary star simulation (one star a red giant with sink particle core, the other  more like a main sequence star, both made of gas particles). This was supplied by Stephen Nielson after a complaint about poor AM conservation on this problem. 

Performance hit might change a bit at higher resolution but it's only a bit of extra storage plus some algebra so can't see it costing too much

Yrisch:
It's sad... :smiling_face_with_tear:  I'll take a look at the paper, but you're surely right. Anyway, having the higher order is a good option for a future PR. 
Yes, this should be fine, and the caching strategy should keep the cost reasonably low.

Yrisch (approved):
Good for me!

